PR #12, opened 2023-12-21
Add Rproj and renv

### LINKED ISSUE #2: Create Rproject and add renv
It's probably a good idea to add `renv` to this repo, so we can manage any packages we might use to create figures.

allyhawkins:
Closes #2 

All I'm doing here is just adding an `.Rproj` file and initiating `renv`. There's nothing in the lock file other than `renv` and the R version yet, but I wanted to get this in separately from a PR where we have figure additions.

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,0 +1,1 @@
+source("renv/activate.R")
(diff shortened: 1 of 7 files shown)


sjspielman (approved):
👍

Issue #1, opened 2023-12-13
Figure 1: Overview of type of data in the portal 


allyhawkins:
The first figure should encompass an overview of the portal and the contents in the portal. This should probably include the following: 

- [ ] Breakdown of the number of samples for each diagnosis. 
- [ ] Breakdown of how many samples of belong to each modality
- [ ] Breakdown of modalities by project - how many projects have bulk, CITE, spatial, etc. 

**How to display this information:**

I think there are a few ways we could show this information. Option 1 would be to just create a basic table where each unique diagnosis or modality is a separate row. If we went this route, we would probably want to create separate tables, so we have Table 1, Table 2, etc. 

Option 2 is to show this information visually. We could do a bar chart where each diagnosis is its own bar, and we show the total number next to the bar. This would mean we would have 50 bars, so we would want to make sure each of the diagnoses are legible. 
Alternatively, we could break this up by project and group the bars by project. However, that would probably get pretty busy since some diagnoses are found across multiple projects, so we would have > 50 bars. 
I did think about a stacked bar chart, where each bar is a project, but then we would need > 50 colors and that seems really bad. If there was a way we could label each part of the bar and have it be legible, that would be a better option than using a single color for each diagnosis. 

We could maybe group diagnosis by similar types (e.g. all blood cancers, all brain tumors, all sarcomas) and have a color for each group. Then we would need to show a larger more detailed table with all the diagnosis in the supplemental. 

**Other things we might include:**

Related to the breakdown of what's in the portal, I think it might be helpful to show an overview of the Portal. I wonder if we want our first figure to be the same as the about page figure and then we have the associated tables that will all get discussed in the same results section. I know we are updating the about page figure, so I'm not sure how that will affect using that here.

allyhawkins:
Related to the first checkbox listed here, we want to have a figure that shows a summary of the different cancer types represented in the portal. At first I was worried that having a figure showing every diagnosis may be iffy, but I actually think I like it. I'm including two options for this panel below:

**Option 1:** 
Here I show all the different diagnoses that we have, ordered by number of samples. 
![no_groupings_disease_bar](https://github.com/AlexsLemonade/scpca-paper-figures/assets/54039191/714fe633-fc88-4933-90c3-a9fb571c659c)


**Option 2:**
[...]

PR #13, opened 2023-12-22
Initial plot showing summary of sample diagnoses

allyhawkins:
Related to #1 

This PR adds in a script that creates an initial version of the sample overview. The resulting plot is 4 panels, one for each diagnosis group. Each panel has a bar chart where each bar is a separate diagnoses and the bar has different patterns based on the disease timing for that sample. Each panel is a different color so all bars in that panel are the same color. 

Some organizational things: 
- I added a folder for `sample-info` and included two tsv files, one with the breakdown of diagnoses groups and the other with the standardized disease timing groups. In the script to create the figure, I read both of these in and join with the sample metadata to properly group samples. 
- I also created a project whitelist that we can use to track which projects are included in the portal, and thus should be included in the figures. I thought about using the project metadata from S3 (which is synced with admin), but that's not always consistent with what's on the portal. I think this is the most straightforward way of making sure the correct projects are listed. 
- I added a `scripts` folder that contains the script used to generate this figure. I also created a `figures/png` folder to hold all of the `png` outputs. I believe we will also need to make `pdfs` so I thought it was a good idea to store those in a separate folder. Reviewers, let me know if you have other preferences for the organization. 

For the actual plotting: 

- I am creating each panel of the plot separately and then using `ggpubr::ggarrange` to combine them into a single figure. I tried `patchwork`, but it was hard to customize, especially in getting the legends to combine together properly, so I went this route instead. I found formatting to be much easier with `ggpubr`. 
- For the patterns, I ended up choosing 5 different patterns available with `ggpattern`. To be honest, it was not easy to get `ggpattern` to cooperate and actually put different patterns in each disease timing slot. The manual one was the only way I got things to work as expected. I also tried to play around with adjusting angles and shapes, but I found the patterns I chose to be the most distinguishable from each other without being over-busy. I'm happy to try other things if reviewers have different preferences. 
- I did try creating one plot and just using `facet_wrap` or `facet_grid`, but I was having trouble getting the patterns to work and the fills to be correct at the same time. Also, it doesn't like making two rows/ cols when there are different values being plotted on that axis for different facets. It only allowed me to make one row or one column, no grid. 
- For the color scheme, I am temporarily using: `palette.colors(n = 5, palette = "R4")[-1]:` `#DF536B` `#61D04F` `#2297E6` `#28E2E5`. The other color schemes that were suggested all have some darker colors and it's hard to see a black pattern. I also tried a white pattern, but in my opinion it wasn't as clear. 

Things I need to fix, but I cannot seem to get working right: 

- The legend title here will not cooperate. I keep trying different ways when I create the legend to do that and nothing has worked. I am altering the legend title when I create the panel, but maybe I should be doing it after I combine the plots? 
- The legend boxes are not black and white. I've tried using `guides( fill = guide_legend(override.aes = list(fill = "white")))` and also with `color` and neither worked. 
- The combined x and y labels show up on the plot in Rstudio, but when you use `ggsave` they save to the black space around the white box for the figure... 

I'm filing this as a draft right now (partially because no one is going to look at it until next year anyways), but also because I want to make sure the code is organized properly and is fully cleaned up before requesting a formal review.


allyhawkins:
I've gone ahead and cleaned up the code here and added comments throughout. I still have not fixed the legend or the x-y axis labels. 
I think we can still start the review process for this, noting that those two things will have to be addressed. I'm going to request @jaclyn-taroni for review first to provide feedback on organization and the plot itself. 

I'm also going to tag @sjspielman or @jashapiro, to see if they have any ideas on how to make the legend black and white, fix the legend title, or make the x/y axis labels visible when using `ggsave`.

scripts/fig1-sample_disease_barchart.R
```diff
@@ -0,0 +1,167 @@
...
+    barplot_panel(diagnosis_plot_df = diagnosis_plot_df,
+                  diagnosis_group = group, 
+                  group_color = group_color)
+  })
+
+# grab one legend from list of plots to use as the common legend
+legend_to_use <- ggpubr::get_legend(all_panels[[1]])
+
+# combine into one plot using a common legend 
+combined_plot <- ggpubr::ggarrange(plotlist = all_panels,
```
jashapiro, line 148: Is there a reason you are using `ggarrange` rather than faceting on group? I feel like that would probably solve the legend color issue.

I think if you sort the diagnoses first by group and then use `scale=free` in the facet, you should be able to get around the label ordering issue you might have been worried about)

figures/pngs/Fig1-sample-summary.png
jaclyn-taroni, line 1: Reiterating a comment we just talked about here: I would expect to see the sample sizes at the "top" (to the right of) the bar because we have different upper limits to the x-axis.

scripts/fig1-sample_disease_barchart.R
allyhawkins: Hmm I did use `scale_x=free` and `scale_y=free`,  but that only worked with creating a single row or single column. I wasn't able to get `facet_wrap` to work as expected, but I can try it again and make sure I use `scale=free`. The issue with faceting was also getting the patterns to work correctly, but I'll play around with it again.

jashapiro: A semi-minimal example that I think gets all the things you want?
```
library(ggplot2)
df <- data.frame(
  group = c("A", "A", "A", "B", "B", "B", "B", "B"),
  diagnosis = c("A1", "A1", "A2", "B1", "B1", "B1", "B2", "B2"),
  timepoint = c("x", "y", "x", "x", "x", "z", "z", "y")
)

ggplot(df, aes(y = diagnosis,  fill = group)) +
  ggpattern::geom_bar_pattern(aes(pattern = timepoint), 
                              color = "black",
                              pattern_color = "black",
                              pattern_fill = "black") +
  facet_wrap(facets = "group", scales = "free") +
  ggpattern::scale_pattern_manual(
    values = c("x" = "none", "y" = "stripe", "z" = "crosshatch")) + 
  guides(fill = "none", 
         pattern = guide_legend(override.aes = list(fill = "white"))) + 
  labs(pattern = "Time Point") + 
  theme(text = element_text(size = 10),
        legend.position = "top",
        legend.key.size = unit(1, 'cm')
  )
```

allyhawkins:
Okay I made some updates using Josh's suggestion in https://github.com/AlexsLemonade/scpca-paper-figures/pull/13#discussion_r1439780095. I must have tried a different variation of faceting when I did it, because this works so much better than whatever I was trying to do in December. 

I've included an updated version of the figure, which includes the sample totals next to each bar. I also added a separate tsv file with the palette used here, in case we need to use it in another figure. 

@jaclyn-taroni if you want to provide any thoughts on aesthetics I can update those and then I'll request @jashapiro for review of the actual code.

palettes/diagnosis-group-palette.tsv
```diff
@@ -0,0 +1,5 @@
+diagnosis_group	color
+Brain and CNS 	#DF536B
+Leukemia	#61D04F
+Sarcoma	#2297E6
+Other solid tumor	#28E2E5
```
sjspielman, line 5: 😞 github new line.

scripts/fig1-sample_disease_barchart.R
```diff
@@ -0,0 +1,121 @@
...
+
+# read in project whitelist and grouping metadata 
+project_whitelist <- readLines(project_whitelist_file)
+diagnosis_groupings_df <- readr::read_tsv(diagnosis_groupings_file) |>
+  dplyr::select(submitted_diagnosis, diagnosis_group)
+disease_timing_df <- readr::read_tsv(disease_timing_file)
+
+# Copy sample metadata 
+local_sample_metadata <- file.path(local_s3_dir, "scpca-sample-metadata.tsv")
+sync_call <- paste('op run -- aws s3 cp', sample_metadata_s3, local_sample_metadata, '--recursive', sep = " ")
```
sjspielman, line 42: Something is funky here - it copies the file _as a directory_. This works for me instead - 

```suggestion
sync_call <- paste('op run -- aws s3 cp', sample_metadata_s3, local_s3_dir, sep = " ")
```

jashapiro: Since I just saw this: `op run` should not be internal to a script: The user should invoke that on the wrapper script as needed.

palettes/diagnosis-group-palette.tsv
allyhawkins, line 5:
```suggestion
Other solid tumor	#28E2E5

```

scripts/fig1-sample_disease_barchart.R
allyhawkins:
```suggestion
sync_call <- paste('aws s3 cp', sample_metadata_s3, local_s3_dir, sep = " ")
```

figures/pngs/Fig1-sample-summary.png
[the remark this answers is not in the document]
jashapiro: > * This is a function of the flipped coordinates, but notice how the patterns in the bars themselves are in the opposite order of the legend. That makes the reader work harder than they need to – we'll want the legend and the patterns in the bars themselves to be in the same order. That might require you to output the legend separately using a different factor order and then paste things back together again.

You should be able do this with adding `reverse = TRUE` to the `guide_legend()` call.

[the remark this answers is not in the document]
jaclyn-taroni: I was hoping there was an easier way, and it seems like there is!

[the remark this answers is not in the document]
jashapiro: Just following up though... you will probably want to change the order of the scale as well, to keep Initial Diagnosis first/on the far left, etc.?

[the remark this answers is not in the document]
jaclyn-taroni: > you will probably want to change the order of the scale as well, to keep Initial Diagnosis first/on the far left, etc.?

Agree.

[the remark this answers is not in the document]
jaclyn-taroni: Another option you have available to you in your quest for larger font sizes would be to break the diagnosis labels into two lines. That might also make things _worse_ but did want to note that possibility.

allyhawkins:
I updated the pattern fill and the legend so that `Initial diagnosis` is still first in the legend and that matches up with the patterns shown in the bars. I also changed the patterns so that `Initial diagnosis` is the blank pattern. 

I updated the size of the fonts, but I'm not sure if it's enough. When I open the actual file it looks good to me, but the preview shown in the diff still looks small. Let me know if I need to bump the size even more. 

The other thing I did was pull out the code for syncing the metadata from S3 into a separate script. I'm storing this script in a folder named `figure_setup`.

sample-info/diagnosis-groupings.tsv
```diff
@@ -0,0 +1,54 @@
+ontology_id	human_readable_value	submitted_diagnosis	diagnosis_group
+MONDO:0016684	Anaplastic astrocytoma	Anaplastic astrocytoma	Brain and CNS
```
jashapiro, line 2: Looking at these diagnosis groups, I was wondering if we wanted to use MONDO terms here too (maybe not for display, but we could include them in the table)? My general thought is that if we can use ontology-based groups, we can more easily justify them.

We could use:
- [`MONDO:0021248`: nervous system neoplasm](http://purl.obolibrary.org/obo/MONDO_0021248) for `Brain and CNS`
- [`MONDO:0002334`: hematopoietic and lymphoid system neoplasm](http://purl.obolibrary.org/obo/MONDO_0002334) for `leukemia`
- [`MONDO:0044334`: `connective and soft tissue neoplasm`](http://purl.obolibrary.org/obo/MONDO_0044334) for sarcomas
And then leave an "other" category.

This would change a few categories, but I am had questions about some of them them anyway: 
- is there a reason you don't have neuroblastoma and retinoblastoma in the Brain and CNS category?
- in the ontology, rhabdoid tumor falls under `connective and soft tissue neoplasm`; I'm not sure if we would want to maintain that...

This makes the categories more unbalanced, which could create a different problem... we might have to revert to making separate bar plots (though I would try using `patchwork` to join them, as I think you can get easier control over placement). 

All of this I might say to push off to a separate PR. The current version strikes me as _good_, so the real question is whether we want to use Ontology-based groupings, and then we might have consequences that flow from that.

jashapiro (approved):
I think this looks good, with some minor changes. 

I also had a more major change thought, about using more formal ontology-based groupings, but I tend to think that should be addressed in a separate discussion/PR, as it could have wider implications beyond this one plot.

sample-info/diagnosis-groupings.tsv
allyhawkins: I don't know that we are planning to include that table that's in the repo in the paper, but we probably should. If we do, then I think having the MONDO term would be a good idea so I'm okay with this change. 

> is there a reason you don't have neuroblastoma and retinoblastoma in the Brain and CNS category?

I was on the fence about Neuroblastoma because most of the cases are peripheral and found in abdomen, adrenal gland, pelvis, etc. There is a separate classification that is CNS Neuroblastoma, which is why I didn't originally include it. But looking at the parent terms for Neuroblastoma in MONDO, you are probably right. I could see an argument for both ways since I don't think it's really a traditional CNS tumor. 

For retinoblastoma, this is just from me thinking of it as a solid tumor that's different from gliomas, even though I guess technically it is a CNS tumor. 

> in the ontology, rhabdoid tumor falls under connective and soft tissue neoplasm; I'm not sure if we would want to maintain that...

I think I would actually use the MONDO: for `sarcoma` instead of the connective and soft tissue. It's more specific to the fact that we are only including sarcomas and not soft tissue which is why Rhabdoid is separate. 

I think right now I will for sure change over Retinoblastoma, but I won't change the other ones. Unless you feel strongly about Neuroblastoma. And then I'll open an issue for deciding if we want to use the MONDO groups or not.

scripts/fig1-sample_disease_barchart.R
```diff
@@ -0,0 +1,115 @@
...
+# create faceted plot, one panel for each diagnosis group
+# diagnosis is on the y axis and number of samples on the x axis
+# bars are patterned based on the disease timing group 
+diagnosis_plot <- ggplot(plot_df, aes(y = diagnosis,  fill = diagnosis_group)) +
+  ggpattern::geom_bar_pattern(aes(pattern = forcats::fct_rev(standardized_disease_timing)), 
+                              color = "black",
+                              pattern_color = "black",
+                              pattern_fill = "black",
+                              pattern_density = 0.04,
+                              pattern_spacing = 0.02)+
```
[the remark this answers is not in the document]
allyhawkins: So I can adjust the spacing or density on all of them to make them more dense. I felt like when I did that, the other ones looked bad and then the circle looked better.

sample-info/diagnosis-groupings.tsv
jashapiro: > I think I would actually use the MONDO: for `sarcoma` instead of the connective and soft tissue.

Makes sense. I think I was leaning toward connective & soft tissue because it was more at the same "level", but that was not a strong sentiment!

scripts/fig1-sample_disease_barchart.R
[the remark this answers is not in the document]
allyhawkins: I went ahead and made the patterns darker so they are easier to see I think.

allyhawkins:
Noting that I'm going to go ahead and merge this and any additional discussion we should have in #20

Issue #14, opened 2024-01-02
Overview of modalities 


allyhawkins:
Related to #1. 
In addition to including a summary of the types of samples (diagnosis, disease timing), we should also include information on the types of modalities available in each project. In particular it would be important to display if samples have libraries that include CITE-seq, have associated bulk data, are multiplexed, or have spatial data. Since samples could have multiple modalities (such as bulk data and CITE-seq), I think a table might be the best approach for showing this: 

| Project ID | Total samples | Bulk | CITE (ADT) | Spatial | Multiplexed | 
| --------- | -------------- | ----- | ----------- | ------- | ----------- |
| SCPCP000001 | 23 | 23 | 0 | 0 | 0 | 

A few things to consider: 
- Do we want to add columns for single-cell vs. single-nuclei? 
- Do we want to summarize the total number of samples or libraries? I think samples make more sense since that's what you download on the portal.

allyhawkins:
For the modalities overview, we are going to present it two ways. 

1. Fig. 1B - Bar plot showing a summary of all libraries broken down by modalities. This plot should have proportion of libraries on the y-axis (ranging from 0 - 100) and then each modality listed on the x-axis (all single-cell/nuclei, bulk, spatial, and CITE). The bars will then be colored based on how many samples are single-cell or single-nuclei. If there are any samples that don't have a matching single-cell/nuclei sample then they will not be included. 

2. Table S1 - This will include a summarized table showing the breakdown of number of samples for each project by type. In addition to the columns shown in https://github.com/AlexsLemonade/scpca-paper-figures/issues/14#issue-2062639488, we will add in a column for diagnosis group and diagnosis. The diagnosis column will contain a semi-colon separated list of all diagnosis found in that project with that diagnosis group.

allyhawkins:
Closed by #17 and #18

PR #17, opened 2024-01-08
Table S1: Overview of modalities 

allyhawkins:
Addresses #14 
Stacked on #13 

This PR adds in the script to generate a table summarizing the total libraries per project and what type of libraries are included. We decided we would like to include a table that has one row per project/ diagnosis grouping combination. So if a project has diagnoses that all fall under the same diagnoses group then there is only one row for that project. If the project contains samples that are from different diagnoses groups then there will be multiple lines. This really only happens when we have `Non-cancerous` samples present in a project. 

For each project/ diagnosis group, I show the total number of libraries and then break that down into single-cell, single-nucleus, bulk, spatial, with CITE, and with cellhashing. Note that if a project contains CITE or cell hashing, the total libraries does not count the CITE/ cell hash libraries twice. To count the total, I pulled out the unique libraries in each project. So if you see that a project contains 60 single-cell and then 60 with CITE-seq, then that just means all 60 single-cell libraries have associated CITE-seq data. 

For bulk and spatial, the library ID is not the same between the single-cell/ single-nuc library and the bulk/spatial library. So although they may be from the same sample (this is not always the case), they will get counted as separate libraries when getting the total. 

Note I did stack this on #13 because I'm using the sample and library metadata from S3 that is downloaded in one of the scripts added in #13.  

One major thing to note is that I am showing the total number of libraries and NOT the total number of samples. I'm curious if others think we should show the total number of samples too?


allyhawkins:
I'm going to request @jaclyn-taroni to just take a look at the table to make sure this is what she expected, and then another scientist for review of the code.

tables/Supplemental-table-1-modality-overview.tsv
```diff
@@ -0,0 +1,21 @@
+scpca_project_id	Diagnosis group	Total number of libraries	Single-cell	Single-nucleus	Bulk RNA	With CITE-seq	Spatial transcriptomics	With cell hashing	Diagnoses
```
jaclyn-taroni, line 1: My only comment is that I would move `Diagnoses` immediately after `Diagnosis group`

jaclyn-taroni (commented):
This mostly looks as I expected. I left one comment about moving a column!

scripts/table-s1-modality-summary.R
```diff
@@ -0,0 +1,109 @@
...
+project_whitelist <- readLines(project_whitelist_file)
+
+# read in groupings
+diagnosis_groupings_df <- readr::read_tsv(diagnosis_groupings_file)
+
+# read in library metadatq and filter to included projects
+library_metadata_df <- readr::read_tsv(library_metadata_file) |> 
+  dplyr::filter(scpca_project_id %in% project_whitelist)
+
+# read in sample metadatq and filter to included projects
```
sjspielman, line 35: metadatq -> metadata

scripts/table-s1-modality-summary.R
```diff
@@ -0,0 +1,109 @@
...
+    # make a new column that summarizes all modality info
+    # options are cell, nucleus, bulk, spatial, cite, or multiplexed
+    # the assigned values will eventually be the column names used in the output
+    modality = dplyr::case_when(
+      seq_unit == "cell"  & !(stringr::str_detect(technology,"CITE")) & !(stringr::str_detect(technology, "cellhash")) ~ "Single-cell",
+      seq_unit == "nucleus"  & !(stringr::str_detect(technology,"CITE")) & !(stringr::str_detect(technology, "cellhash")) ~ "Single-nucleus",
+      seq_unit == "bulk" ~ "Bulk RNA",
+      seq_unit == "spot" ~ "Spatial transcriptomics",
+      stringr::str_detect(technology,"CITE") ~ "With CITE-seq",
+      stringr::str_detect(technology,"cellhash") ~ "With cell hashing"
```
sjspielman, line 74: If you move these earlier, you can get rid of some the `!str_detect` code

scripts/table-s1-modality-summary.R
```diff
@@ -0,0 +1,109 @@
...
+  dplyr::group_by(scpca_project_id, diagnosis_group) |> 
+  dplyr::mutate(diagnosis = paste(unique(diagnosis), collapse = ";")) |>
+  # make sure that diagnosis now gets retained throughout creation of the table
+  dplyr::group_by(scpca_project_id, diagnosis_group, diagnosis) |> 
+  dplyr::count(scpca_project_id, modality) |> 
+  # create a column for each modality that contains the total number of libraries
+  tidyr::pivot_wider(names_from = modality, values_from = n) |> 
+  dplyr::mutate(
+    # replace all NA with 0
+    across(everything(), ~ tidyr::replace_na(.x,0))
```
sjspielman, line 93:
```suggestion
    across(everything(), ~ tidyr::replace_na(.x, 0))
```

scripts/table-s1-modality-summary.R
```diff
@@ -0,0 +1,109 @@
...
+    "Total number of libraries", 
+    "Single-cell", 
+    "Single-nucleus", 
+    .after = 1
+    ) 
+
+
+# export table 
+readr::write_tsv(modality_counts_df, output_table_file)
+
```
sjspielman, line 109: new line

sjspielman (commented):
This is _extremely not_ a full review, just a quick scroll through to get a sense of the code and small things I saw during said scroll.

scripts/table-s1-modality-summary.R
```diff
@@ -0,0 +1,109 @@
...
+  dplyr::select(
+    scpca_project_id, scpca_library_id, modality, diagnosis, diagnosis_group
+  ) |> 
+  # for each project, group by diagnosis group, count the total number of libraries
+  # and then summarize the diagnoses included 
+  dplyr::group_by(scpca_project_id, diagnosis_group) |> 
+  dplyr::mutate(diagnosis = paste(unique(diagnosis), collapse = ";")) |>
+  # make sure that diagnosis now gets retained throughout creation of the table
+  dplyr::group_by(scpca_project_id, diagnosis_group, diagnosis) |> 
+  dplyr::count(scpca_project_id, modality) |>
```
sjspielman, line 88: You can just count here, which will also do (and then get rid of, which is good!) the grouping. The outcome looks the same to me in the end, unless I've missed something?

```suggestion
  dplyr::count(scpca_project_id, diagnosis_group, diagnosis, modality) |> 
```

sjspielman (commented):
Overall the result table looks correct to me, and this is mostly there! See inline comments..

> One major thing to note is that I am showing the total number of libraries and NOT the total number of samples. I'm curious if others think we should show the total number of samples too?

I do think this is worth showing because libraries:samples is not 1:1, for several reasons (multiplexing, and as you noted in the PR, "So although they may be from the same sample (this is not always the case), they will get counted as separate libraries when getting the total.")

People _love_ statistics like this (samples and libraries), and we're already in the supplemental so might as well have two side-by-side tables. I think the code here is small enough to just tack that extra code on without _needing_ a function, but I'm not entirely sure I'm right...

We'll also want to keep these items in mind for captions. Speaking of which, what/when is our plan to write those? Waiting for the paper itself, or should we get going on them alongside figure/table creation?

allyhawkins:
> People love statistics like this (samples and libraries), and we're already in the supplemental so might as well have two side-by-side tables. I think the code here is small enough to just tack that extra code on without needing a function, but I'm not entirely sure I'm right...

I'm good with adding a column that has total samples to this existing table. But I'm not sure we want to have a separate table. I'm on the fence on this so I'll add it in and go from there. 

> We'll also want to keep these items in mind for captions. Speaking of which, what/when is our plan to write those? Waiting for the paper itself, or should we get going on them alongside figure/table creation?

I was thinking figure captions/ legends would be separate.

scripts/table-s1-modality-summary.R
```diff
@@ -0,0 +1,109 @@
...
+    # options are cell, nucleus, bulk, spatial, cite, or multiplexed
+    # the assigned values will eventually be the column names used in the output
+    modality = dplyr::case_when(
+      seq_unit == "bulk" ~ "Bulk RNA",
+      seq_unit == "spot" ~ "Spatial transcriptomics",
+      stringr::str_detect(technology,"CITE") ~ "With CITE-seq",
+      stringr::str_detect(technology,"cellhash") ~ "With cell hashing",
+      seq_unit == "cell" ~ "Single-cell",
+      seq_unit == "nucleus" ~ "Single-nucleus",
+    )
```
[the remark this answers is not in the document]
allyhawkins: Yea I'm using the library metadata file that has one row per run.

sjspielman:
> I'm good with adding a column that has total samples to this existing table.

Seems easier!!

allyhawkins:
Okay I added in a column for total sample count and implemented your other suggestions @sjspielman. 

I think we should decide if we like just adding a column with the total samples, but then noting in the caption that the other columns show the total in libraries. Alternatively, we could create an entirely separate table where we show the breakdown of samples rather than libraries. I do think the separate table would be less confusing then showing one column with sample counts and all other columns with library counts. If we do that though I think we need another PR. I don't think the data wrangling will be exactly the same. Thoughts @jaclyn-taroni?

scripts/table-s1-modality-summary.R
```diff
@@ -0,0 +1,109 @@
...
+  dplyr::select(scpca_project_id, scpca_sample_id, diagnosis) |> 
+  # add diagnosis groupings
+  dplyr::left_join(diagnosis_groupings_df, by = c("diagnosis" = "submitted_diagnosis"))
+
+# Create table -----------------------------------------------------------------
+
+# create a list of all the modality values
+# we will use this to total the number of libraries for a project across all modalities 
+# these will also correspond to the column names in the final table that contain the total for each modality 
+modality_columns <- c("Single-cell", "Single-nucleus", "Bulk RNA", "Spatial transcriptomics", "With CITE-seq", "With cell hashing")
```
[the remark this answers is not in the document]
sjspielman: Sticking with `CITE-seq`. See https://github.com/AlexsLemonade/scpca-paper-figures/pull/18#issuecomment-1883480435

sjspielman:
Hm, now I do wonder if it is confusing that the table modality columns show libraries. But, I think that confusion is easily addressed with little superscripts for those columns in the final table itself, with an associated footnote or caption section that says "these are library, not sample, counts."

allyhawkins:
> Hm, now I do wonder if it is confusing that the table modality columns show libraries, *given the additional # samples column. But, I think that confusion is easily addressed with little superscripts for those columns in the final table itself, with an associated footnote or caption section that says "these are library, not sample, counts."

I'm good with the idea of incorporating superscripts or some other indicator that they are libraries. What about adding (L) or (S) to the column titles to indicate which count is shown? Most of them will be L except for the total samples. Then we can describe the (L) and (S) in the caption?

sjspielman:
Yeah, I think either of those ideas are fine! The specific format (superscript vs indicator string) may end up depending on the journal, so we can maybe decide on the right approach once we finalize our submission plan?

allyhawkins:
For now, I just added the S and L indicators to the column names. If we need to change the formatting for the journal we can do that at a later time.

PR #18, opened 2024-01-08
Fig 1B: Modality overview bar chart

allyhawkins:
Addresses #14 
Stacked on #13 

This PR adds in a bar chart that summarizes the number of samples with multiple modalities. In our original discussion, we wanted to show a column with the total number of samples on the portal, colored by whether or not the sample is single-cell or single-nuclei. Then, we want to show the total number of samples that contain an additional modality (bulk, spatial, cite, or cell hashing). These bars are also colored by single-cell vs. single-nuclei. 

A few important notes: 

- I wasn't sure if we wanted to show the total number of samples vs. the total number of libraries. The reason I'm showing samples here is that we group everything by sample on the portal, and we don't publish the total number of libraries per project or allow users to download a single library. Additionally, we have different library IDs for a single-cell library vs. bulk library, even though they come from the same sample. I think the purpose of this plot is to show how many of the samples on the portal have additional information obtained from bulk or another modality. So samples seemed to make more sense to me. 
- I decided to "facet" this plot in a way that separates the `Single suspension` or all samples from samples with additional modalities. The bar showing `All samples` is a summary of every sample available in the portal. Then the next "facet" shows how many samples indicated in the `All samples` bar have associated bulk, CITE, etc. I'm curious what people think of faceting like this? I was looking for a way to distinguish to readers that the first bar has every single sample regardless of additional modality. 
- I think in my original discussion with @jaclyn-taroni, we wanted to show the percentage on the y-axis so that the y-axis is 0-100 with the `All samples` bar going up to 100. Then the height of each additional modality bar would be the percentage of all samples that have that modality. I had a few issues getting `ggplot` to plot the height as the percentage but then fill the bars based on cell/ nuclei. So I just used the total number of samples on the y-axis. I don't think the overall look of the plot would change, just the axis range would be 0-100 instead of 0-500. Are we okay with this instead?
- For the colors, I chose two colors from `Set2` that I think looked good together. I added a new palette for these colors, but I'm happy to change them.  

Again, I'm going to ask @jaclyn-taroni to look at the plot and provide any feedback on plot changes, and then I'll request review of the code.


figures/pngs/Fig1B-modality-summary.png
jaclyn-taroni, line 1: I might reorder these bars – I expect it might be helpful to put `With CITE-seq` and `With cell hashing` next to each other when you go to write about it in the text.

I would also try adding labels with the sample number at the top ([example](https://r-graphics.org/recipe-bar-graph-labels#fig:FIG-BAR-LABEL-STACKED)) or middle ([example](https://r-graphics.org/recipe-bar-graph-labels#fig:FIG-BAR-LABEL-STACKED-MIDDLE)) of each bar.

jaclyn-taroni (commented):
I agree with showing samples, not libraries. I'm not sure I have a better suggestion about layout than what you've done, so that is an area where it might be helpful to hear from others. I left a couple of comments on the plot file itself.

figures/pngs/Fig1B-modality-summary.png
allyhawkins: I updated the plot to include labels in the middle of each bar and also moved CITE and cell hashing to be next to each other. The updated plot is in https://github.com/AlexsLemonade/scpca-paper-figures/pull/18/commits/13ce647b4893d3b71a44a5943384261022d6122b.

allyhawkins:
I'm going to tag in @jashapiro and @sjspielman to see if you have any feedback on the aesthetics of this plot. See https://github.com/AlexsLemonade/scpca-paper-figures/pull/18#issue-2071226572 for the description. 
You can find the most recent version here: https://github.com/AlexsLemonade/scpca-paper-figures/blob/allyhawkins/modality-bar-chart/figures/pngs/Fig1B-modality-summary.png

scripts/fig1b-modality_barchart.R
```diff
@@ -0,0 +1,138 @@
...
+  theme(legend.position = c(.8,.9),
+        legend.direction = "horizontal",
+        legend.box.background = element_rect(colour = "black"),
+        text = element_text(size = 14),
+        axis.text.x = element_text(angle = 45, hjust = 1)) +
+  guides(fill = guide_legend(title.position = "top", title.hjust = 0.5))
+
+ggsave(output_plot_file)
+
+
```
sjspielman, line 138: 😿

sjspielman:
Aesthetic looks good to me!! 🎨  

The only thought I have (which mirrors one of the comments I left in #17) is whether `with CITE-seq` or `with ADT` is the right term to be using throughout the paper. I keep going back and forth... CITE-seq is more recognizable as a term, but ADT is more technically correct. In the portal, we use a `CITE-seq` filter, and docs use both CITE-seq and ADT terms in different settings where appropriate (e.g, https://scpca.readthedocs.io/en/stable/sce_file_contents.html#additional-singlecellexperiment-components-for-cite-seq-libraries-with-adt-tags)

I don't know the answer here, so hearing other people's thoughts will be good!

scripts/fig1b-modality_barchart.R
sjspielman: ^ the github monster needs a new line

allyhawkins:
> The only thought I have (which mirrors one of the comments I left in https://github.com/AlexsLemonade/scpca-paper-figures/pull/17) is whether with CITE-seq or with ADT is the right term to be using throughout the paper. I keep going back and forth... CITE-seq is more recognizable as a term, but ADT is more technically correct. In the portal, we use a CITE-seq filter, and docs use both CITE-seq and ADT terms in different settings where appropriate (e.g, https://scpca.readthedocs.io/en/stable/sce_file_contents.html#additional-singlecellexperiment-components-for-cite-seq-libraries-with-adt-tags)

The reason I'm using `with CITE-seq` here is because on the portal we use `CITE-seq` throughout, so I want to mirror the language that we are showing people on the portal itself.

scripts/fig1b-modality_barchart.R
```diff
@@ -0,0 +1,138 @@
...
+  dplyr::group_by(scpca_sample_id, seq_unit) |>
+  dplyr::summarise(modality = paste(modality, collapse = ";")) |> 
+  # split any lists of modalities so we have one row per sample per modality
+  dplyr::mutate(modality = stringr::str_split(modality, ";")) |> 
+  tidyr::unnest(modality) |>
+  dplyr::group_by(modality) |> 
+  # get the total for each modality to use for specifying order in plot
+  dplyr::add_tally(name = "total_per_modality") |>
+  # add a column to help pull out additional modalities into its own facet 
+  dplyr::mutate(additional_mods = ifelse(modality == "Single suspension", "All samples", "Samples with additional modalities"),
```
sjspielman, line 94: Can you generally update the spacing here, with `additional_mods =` on new line, with `ifelse` maybe spanning a couple lines?

scripts/fig1b-modality_barchart.R
```diff
@@ -0,0 +1,138 @@
...
+                modality = as.factor(modality),
```
sjspielman, line 95: This can go; running the `forcats` next should implicitly make it a factor first.
```suggestion
```

scripts/fig1b-modality_barchart.R
```diff
@@ -0,0 +1,138 @@
...
+                # set custom order
+                # with cite and with cell hash should be together 
+                modality = forcats::fct_relevel(modality,
+                                                "Single suspension",
+                                                "Bulk",
+                                                "Spatial transcriptomics",
+                                                "With CITE-seq",
+                                                "With cell hashing"
```
sjspielman, line 103: I might define this `modality_order` vector towards the top of the script for future changes, which probably aren't likely, but can't hurt.

scripts/fig1b-modality_barchart.R
```diff
@@ -0,0 +1,138 @@
...
+# the bars are colored by what suspension type that modality is composed of
+ggplot(filtered_modality_df, aes(x = modality, fill = seq_unit)) +
+  geom_bar(stat = "count", color = "black") +
+  # split the graph to designate all samples vs. those with additional modalities
+  facet_grid(cols = vars(additional_mods),
+             scales = "free",
+             space = "free") +
+  theme_classic() + 
+  # add label to middle of each bar 
+  geom_text(aes(label = after_stat(count)), stat = "count", position = position_stack(vjust = 0.5)) +
```
sjspielman, line 122: Space this out onto a few lines, it's a little long

scripts/fig1b-modality_barchart.R
```diff
@@ -0,0 +1,138 @@
...
+  labs(
+    x = "",
+    y = "Number of samples",
+    fill = "Suspension type"
+  ) + 
+  scale_fill_manual(values= suspension_colors) +
+  theme(legend.position = c(.8,.9),
+        legend.direction = "horizontal",
+        legend.box.background = element_rect(colour = "black"),
```
sjspielman, line 131: 🇺🇸 
```suggestion
        legend.box.background = element_rect(color = "black"),
```

scripts/fig1b-modality_barchart.R
sjspielman: ^^^

sjspielman: i guess there must be spaces on those final lines which is why github is being so difficult?

scripts/fig1b-modality_barchart.R
```diff
@@ -0,0 +1,138 @@
...
+        text = element_text(size = 14),
+        axis.text.x = element_text(angle = 45, hjust = 1)) +
+  guides(fill = guide_legend(title.position = "top", title.hjust = 0.5))
+
+ggsave(output_plot_file)
```
sjspielman, line 136: Can we specify a width/height here to ensure reproducibility? The default is 7x7, so maybe just pop those in since it looks fine. IMO this is a good practice to ensure that the chosen text sizes will definitely work!

sjspielman (approved):
LGTM! I left a few very small comments, mostly spacing. I don't feel that I need to see this again, unless you want me to have another look!

PR #27, opened 2024-01-11
Fig 2B: Simplified and mini QC report plots

### LINKED ISSUE #6: Figure 2B: Mini QC report plots
After explaining the overview of the figure, I think it will be beneficial to include some plots that show each of the steps taken when processing each library. We include a lot of this in our main QC report. This figure should include an example of each of the following plots that are found in our QC report: 

- [ ] Knee plot showing which cells passed filtering out empty droplets 
- [ ] Summary of cell metrics showing the total umi/ number of genes/ mitochondrial content for each cell
- [ ] miQC model plot
- [ ] Plot showing removal of low quality cells (cells colored by keep/ remove)
- [ ] UMAP colored by number of genes detected OR do we show clusters...? Or both? 

To do this, we will need to find an example library that we like to use for re-creating each of these plots.

allyhawkins:
Closes #6 

This PR adds a summarized and simplified version of the QC report plots as a figure panel. We want to include "mini" versions of each of the plots included in the main QC report, but I wasn't sure how mini we wanted to go. Figure 2 will be the scpca-nf overview and the panel I included here, so I think if it takes up half a page that's okay and we can always downsize it. 

At first, I tried to just shrink the plots on illustrator, but then I realized that in order to both shrink the plot and keep the fonts legible I needed to have a pdf version of the plot, not an image file. We don't output that, so I started to write a script to create individual pdf versions of each plot and then save them to edit in illustrator, but then I remembered that I'm much better at figuring out R than I am at illustrator. So this script creates all 6 plots, but I just simplified them and made sure they were uniform. Then I used patchwork to combine them into one plot and saved them as a png. 

For these simplified figures, I made the following decisions: 

- I removed the axis ticks and labels. For these plots, we don't particularly care about the exact numbers, we just want to see what information you can obtain from viewing these plots. 
- I made the legends smaller and moved them to the same upper corner for the knee plot, read metrics, and miQC plot. In doing this, I removed the legend title and the labels on the legend. I was unsure whether we wanted to keep the legend title, but I do think it will be hard to make the legend label legible? I'm curious what others think here. 
- I removed the grids from all of the plots and am using `theme_minimal()`. But I did add a border around the plot so each plot would appear distinct.
- For the UMAPs, I removed the legend all together... I'm not totally sure if this was the right call, but for those plots we included the legend on the side or bottom, so having the legend there messed up the prettiness of everything lining up. I think for the individual UMAP we could add a legend inside the box, since we know where the points will be, but not sure what to do about the faceted UMAP. 
- For the faceted plot where we show the top HVGs, I decided to only show 4, not 12. 12 mini UMAPs seemed like a lot. Also, the top HVG doesn't have a gene symbol which was hard to read in the facet, so for show purposes, I'm showing the 2-5 top HVG. 

In addition to the plots, I also: 

- Started a script to use for downloading any data files from S3 that we may need for future figures. Right now it explicitly downloads the files for SCPCS000001, but I'm sure that will change as we include more plots. 
- I updated the readme's to include mention of this figure and a description of the two new scripts. 
- I noticed that I couldn't run any of these scripts from the command line, because I wasn't loading the Rproj, so I added a line to do that to all of the scripts. 

Once we know the size of the figures that we need to submit (based on the journal), we should be able to go in and edit the size of the image and font sizes appropriately.
Again, I'm requesting @jaclyn-taroni to look at aesthetics and then @jashapiro for aesthetics and code.


figures/pngs/Fig2B-mini-qc-plots.png
jaclyn-taroni, line 1: The dotted line in the keep/remove plot is somewhat distracting without more information. Other than that, this looks fine to me for now. Unfortunately, I'm not sure we'll be able to make a final.final determination until you lay it out 😬

allyhawkins: By lay it out, you mean combine this with the overview plot that will be Fig. 2A?

scripts/fig_2b-qc_plots.R
```diff
@@ -0,0 +1,278 @@
...
+                  cell_metrics_plot,
+                  miQC_plot,
+                  filtered_plot,
+                  umap_plot, 
+                  hvg_plot)
+patchwork::wrap_plots(plot_list, ncol = 3) &
+  theme(text = element_text(size = 10))
+
+# save files 
+ggsave(output_plot_file, width = 8.5, height = 5.5, units = "in")
```
jashapiro, line 278: I'd just be explicit here, since it is part of a script.
```suggestion
combined_plot <- patchwork::wrap_plots(plot_list, ncol = 3) &
  theme(text = element_text(size = 10))

# save files 
ggsave(output_plot_file, plot = combined_plot, width = 8.5, height = 5.5, units = "in")
```

scripts/figure_setup/sync-data-files.R
```diff
@@ -0,0 +1,12 @@
...
+# All files will be saved in a folder named `s3_files`
+
+renv::load()
+
+# sync results files for SCPCP000001
+qc_s3_files <- "s3://nextflow-ccdl-results/scpca-prod/results/SCPCP000001/SCPCS000001"
+qc_local_files <- here::here("s3_files", "SCPCS000001")
+fs::dir_create(qc_local_files)
+
+sync_call <- paste('aws s3 cp', qc_s3_files, qc_local_files, "--exclude '*'", "--include '*.rds'", "--recursive", sep = " ")
```
jashapiro, line 11: Totally minor, but as things get complex, I'd tend to switch over to `glue` (and quoting paths is never bad).
```suggestion
sync_call <- glue::glue("aws s3 cp '{qc_s3_files}' '{qc_local_files}' --exclude '*' --include '*.rds' --recursive")
```

jashapiro (approved):
LGTM, with a few tiny comments.

scripts/fig_1a-sample_disease_barchart.R
```diff
@@ -2,6 +2,8 @@
 # The resulting plot is faceted by the diagnoses group 
 # Individual bars have different patterns based on the disease timing 
 
+renv::load()
```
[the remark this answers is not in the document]
allyhawkins: Okay, maybe it was that I was running it in the `scripts` directory, but either way I think it would be helpful to leave this as a safety net.

allyhawkins:
> The dotted line in the keep/remove plot is somewhat distracting without more information. Other than that, this looks fine to me for now. Unfortunately, I'm not sure we'll be able to make a final.final determination until you lay it out 😬

@jashapiro any thoughts on the dotted line? I think we could explain it in the legend, but since we use it to indicate the minimum cutoff in the actual reports, I feel like we need it. Thoughts?

jashapiro:
> > The dotted line in the keep/remove plot is somewhat distracting without more information. Other than that, this looks fine to me for now. Unfortunately, I'm not sure we'll be able to make a final.final determination until you lay it out 😬
> 
> @jashapiro any thoughts on the dotted line? I think we could explain it in the legend, but since we use it to indicate the minimum cutoff in the actual reports, I feel like we need it. Thoughts?

I don't have strong thoughts, but I think we can leave it off for aesthetics here: I don't feel like these plots have to perfectly represent the plots in the report, just give a general indicator of what would be found there.

scripts/fig_2b-qc_plots.R
```diff
@@ -0,0 +1,278 @@
...
+theme_set(
+  theme_classic() +
+    theme(
+      #plot.margin = margin(rep(20, 4)),
+      strip.background = element_rect(fill = "transparent"),
+      # no background box for legend
+      legend.background = element_blank(),
+      # no axis ticks or labels
+      axis.ticks = element_blank(),
+      axis.text = element_blank(),
```
jashapiro, line 20: Minor thought because some doubled line drawing was bothering me a touch.
```suggestion
      axis.line = element_blank(),
      axis.ticks = element_blank(),
      axis.text = element_blank(),
```

Issue #4, opened 2023-12-13
Supplemental Figure 1: Benchmarking of Alevin-fry vs. Cell Ranger


allyhawkins:
We should mention some of the benchmarking that we did to compare results between libraries processed with Cell Ranger and Alevin-fry to help explain why we built our workflow with Alevin-fry. This probably does not need to be a main figure and can be a supplemental figure (unless others have different opinions). 

This figure should include: 

- [ ] Comparison of run time and memory usage between both methods 
- [ ] Comparison of total UMI/cell 
- [ ] Comparison of total genes/cell
- [ ] Comparison of mean gene expression 

All of the plots except for the run time/ memory usage exist as plots in the [FAQ: Why did we use Alevin-fry for processing?](https://scpca.readthedocs.io/en/stable/faq.html#why-did-we-use-alevin-fry-for-processing). We should polish those figures and move the code for generating those figures to this repo.

allyhawkins:
So I started working on this, in particular, I was going back through the previous time and memory comparison we used to compare Cell Ranger and Alevin-fry runs. The original benchmarking results can be found in https://github.com/AlexsLemonade/alsf-scpca/blob/main/analysis/quantifier-comparisons/03-memory_time_comparison.Rmd. 

I have a folder of trace files that were generated during benchmarking that were used in that comparison that I can use, but there are a few caveats. 

- We used Cell Ranger v6, which does not retain introns for single-cell samples. This did come out after we did our original benchmarking, but do we want to re-run with Cell Ranger v7 for better comparisons? I don't think this is imperative, but I did want to note the difference in the Cell Ranger versions. Note that in the images we have in the FAQ's, we use Cell Ranger v6. 
- For Alevin-fry we used a reference without intronic reads for single-cell samples in that particular analysis. But we did other later benchmarking runs with the index we use now, which we used to generate the FAQs. The problem is that I don't have the trace files for those, so I can't get the time or memory information. I will, at minimum, need to re-generate the trace files for the Alevin-fry runs. 
- For all of these runs, we used Ensembl 103 for the indices. If we don't want to re-run Cell Ranger, then I will need to use that older index for the re-runs of Alevin-fry. But if we want to be more up-to-date with what's on the portal, we should use Ensembl 104, which would mean re-runs of the benchmarking runs for both Cell Ranger and Alevin-fry that we want to use to create the supplemental figure.

Here's what I think we should do here: 

[...]

PR #28, opened 2024-01-12
Supplemental figure 1A: benchmarking time and memory usage

allyhawkins:
Related to #4 

This PR adds the first part of the benchmarking figure and includes a panel that looks at total run time and peak memory usage between Cell Ranger and Alevin-fry. To create these figures, I re-ran 6 samples (3 cell, 3 nuclei) through both methods using the same index, Ensembl v104. I added the `-with-trace` command to get the trace file with run time and memory information. 

- I added the trace files that I'm getting this information from to the repo. They're really small so it didn't make sense to put on S3, and I think we should keep them somewhere central so we don't lose them. There's one for the run of Cell Ranger and one for the run of Alevin-fry. 
- For wrangling the trace files, I took a lot of the code from our original benchmarking notebook (https://github.com/AlexsLemonade/alsf-scpca/blob/main/analysis/quantifier-comparisons/03-memory_time_comparison.Rmd). I simplified it and cleaned it up, because wow that code is _wild_. 
- I tried a few different approaches for the plot itself and I ended on a box plot with a sina plot... I know there is a lot of dislike towards boxplots, but with this few samples, I personally found it more visually appealing. I tried just a sina plot with a median line and a sina with violin and the violins just looked odd to me. So I'm curious what others think about the look here. 
- I chose to keep all single-cell and all single-nuclei together, instead of separating them. It just looked so sparse with only 3 samples per condition and they all are similar anyways. 
- These are my favorite colors I've ever used in a plot, but I know it might be hard to see the dots 😢 I can change to have two lighter colors, if people are having trouble. I am planning on using whatever colors we pick here in the rest of the benchmarking figure. 
- I also am wondering if it's okay to use the same type of plot here. I know the axes are different and they are technically measuring different things, but these different things are pretty related. 

Same deal as before, requesting @jaclyn-taroni for aesthetics and @jashapiro for aesthetics and code.


scripts/fig_s1a-memory-time-comparison.R
```diff
@@ -0,0 +1,128 @@
...
+palette <- readr::read_tsv(palette_file)
+
+# get list of all colors 
+method_colors <- palette$color |> 
+  purrr::set_names(palette$method)
+
+
+# compare run time in minutes
+time_plot <- ggplot(grouped_log_df, aes(x = method, y = total_time, fill = method)) +
+  geom_boxplot(outlier.shape = NA) +
```
jashapiro, line 96: For the plot I was thinking something like: 
```
ggplot(grouped_log_df, aes(x = run_id, y = total_time, fill = method) + 
 geom_bar(position="dodge")
```

figures/pngs/FigS1A-time-memory-benchmarking.png
jaclyn-taroni, line 1: > With only 5 points, I wonder if we want a whole box plot, or maybe just showing the mean and the points. Part of me also wants to connect the points between samples run under each method, but I expect that will be busy.

I agree with no box plot for this number of samples and will add: what if each sample got its own shape as an alternative to connecting points?

allyhawkins: > Edited to add: To potentially head off a question here, I don't think the colors are really adding that much information. Edited, again: I suppose it depends on the formatting for the rest of the benchmarking figures.

If we keep the same format of the rest of the figures that we have in the [FAQ](https://scpca.readthedocs.io/en/stable/faq.html#why-did-we-use-alevin-fry-for-processing), then the colors will be used. 

> I agree with no box plot for this number of samples and will add: what if each sample got its own shape as an alternative to connecting points?

I like the idea of trying different shapes!

jashapiro:
I like the shapes and bars, I think... I'd remove the SCPCR legends from all of them though. And for the bars I would order by time?

jaclyn-taroni:
If we do the bars, that opens up the possibility to facet by single-cell vs. single-nuclei.

allyhawkins:
What about something like this? I ordered by time and memory and then faceted by seq unit. @jaclyn-taroni @jashapiro 

image

allyhawkins:
Okay, I've adjusted the code to now have the bar plot pictured above with some slight aesthetic fixes. @jaclyn-taroni let me know if you have any additional comments and @jashapiro this should be ready for review now.

jashapiro (approved):
LGTM

palettes/method-palette.tsv
```diff
@@ -0,0 +1,3 @@
+method	color
+Alevin-fry	#00274C
+Cell Ranger	#FFCB05
```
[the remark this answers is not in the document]
allyhawkins: No way.

PR #30, opened 2024-01-17
Standardize scripts and figure file naming

allyhawkins:
Inspired by a recent discussion on `_` vs. `-`, I was trying to be consistent in the filenames in this repo. But I did it backwards, which is really going to bother me moving forward. So all I did here was update the script file names and the figure files to use `fig-1_description-of-figure.png`. This way moving forward I can use the proper implementation of Jenny Bryan's file naming rules.


scripts/README.md
```diff
@@ -29,14 +29,17 @@ op run -- Rscript sync-data-files.R
 
 The following scripts can be used to generate figures and tables:
 
-1. `fig_1a-sample_disease_barchart.R`: This script is used to generate Figure 1A, which includes a summary of the types of diagnoses found in the ScPCA portal.
+1. `fig-1a_sample-disease-barchart.R`: This script is used to generate Figure 1A, which includes a summary of the types of diagnoses found in the ScPCA portal.
```
jashapiro, line 32: Do you want to do this as `fig1a_sample-disease-barchart.R` (no `-` after `fig`) to be more consistent with the names you give to the figures themselves?

allyhawkins: Good point, I'll make that change!

jashapiro (approved):
LGTM!

PR #31, opened 2024-01-17
Add sample download illustration

### LINKED ISSUE #16: Fig. 2C: Sample download structure
For figure 2, we are focusing on the overview of the workflow, including what is provided as output from the workflow. Because of this, it might be useful to show the illustration showing sample downloads. This way users can connect the output from `scpca-nf` to the data available on the portal. I think we should just be able to take the sample download from the docs illustrations and insert it here.

allyhawkins:
Closes #16 

This PR just adds in the illustration that we have in the docs for downloading an individual sample. We will have to combine this figure with the scpca-nf overview and mini QC plots in illustrator to actually finish figure 2. But for now, I'm just adding this to the repo as an individual png for use later.


sjspielman:
😂 ✅  Screenshot 2024-01-17 at 5 18 10 PM

sjspielman (approved):
🐕

PR #32, opened 2024-01-17
Add script to generate figures and tables

### LINKED ISSUE #23: Create a script to generate all figures and tables
We should create a simple bash script that can be run to generate all the figures and tables that are generated by scripts in the repo. 
We can then use this to create a GHA as mentioned in #21.

sjspielman:
Closes #23 

This PR adds a bash script to generate figures and tables, as well as associated documentation. The script was somewhat inspired by organizational choices in our dear old friend, the [OpenPBTA figure generation script](https://github.com/AlexsLemonade/OpenPBTA-analysis/blob/master/figures/generate-figures.sh).

The script is called `generate-figures-tables.sh` and lives in the top-level directory. I ran the script to ensure it worked, and indeed it did! But on the way I had some `renv` problems, and it seemed that `ggforce` hadn't yet been snapshotted into the lockfile, so that is done here as well. The snapshot added some BioC refs which seems fine. 
I also included a bonus line to purge `Rplots.pdf`, which gets created all the time and nobody really knows why (well, I certainly have never known why and gave up trying to figure it out a long time ago - it's something to do with ggplot innards...) 

Now, when we write new scripts, we'll want to make sure to add them into this bash script as we go.

Note also my wording that "currently, only data lab members can do the S3 things." I assume when the paper goes out we'll make that bucket public, and we can change up some of that wording. But, since this repo is already public, I figured it's the right language for now. We probably don't _need_ it though, so let me know what you think.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,6 +52,14 @@ Rscript scripts/figure_setup/sync-metadata.R
 Rscript scripts/figure_setup/sync-data-files.R
 ```
 
+If you have setup `1Password` to handle your AWS credentials, you will need to prefix those lines with `op run --`:
+
+```sh
+op run -- Rscript scripts/figure_setup/sync-metadata.R
+op run -- Rscript scripts/figure_setup/sync-data-files.R
+```
+
+
 ## Sample info
 
 The `sample_info` folder contains metadata files used to create figures and tables.
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ This repo contains the figures and tables included in the ScPCA manuscript.
 
 ## Summary of figures and tables
 
-Below is a summary of all figures in the paper.
+Below is a summary of all figures and tables in the paper.
 
 **Figure 1**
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -35,7 +35,22 @@ B. Simplified versions of plots in the main QC report.
...
+bash generate-figures-tables.sh
+```
+
+Note that this script assumes that the `s3_files` directory has been populated with relevant data files.
+These files can be obtained by first running the figure setup scripts, which currently require S3 bucket access as a Data Lab member.
+Setup scripts can be run as:
+
+```sh
+Rscript scripts/figure_setup/sync-metadata.R
+Rscript scripts/figure_setup/sync-data-files.R
```
allyhawkins, line 52: Should we add `op run --` here?

generate-figures-tables.sh
```diff
@@ -0,0 +1,71 @@
...
+#
+# Rscript scripts/figure_setup/sync-data-files.R
+# Rscript scripts/figure_setup/sync-metadata.R
+#
+# This script can be run as:
+# bash generate-figures-tables.sh
+
+# enviroment settings
+set -e
+set -o pipefail
```
allyhawkins, line 17:
```suggestion
set -euo pipefail
```

allyhawkins (approved):
Thanks for doing this. This looks good, just some minor comments. I don't need to see this again.

README.md
sjspielman: I think if we're keeping the S3 text for now as "you need to be a data lab member," then yeah it's fine to add this. 
I'll open another issue "before we go public, clean up this README to be more public-facing".

PR #35, opened 2024-01-19
Update S1A to use new method palette

allyhawkins:
This just updates Fig S1A to use the new method palette that we implemented in #29. I am very sad about it, but we should be consistent with our colors in this figure. 

I also added the new script from #29 to the bash script for generating all figures and tables.


jashapiro (approved):
go away blue 😢

Issue #8, opened 2023-12-13
Figure S2: Working with data with additional modalities


allyhawkins:
Here, we want to discuss samples with additional modalities. I think for this figure, we want to show the additional information that is included in a download for libraries that also have CITE-seq and multiplexed. For those that have spatial or bulk, I think a text note about the download format will be sufficient. 

We should show examples of the plots that are included with CITE-seq samples: 

- [ ] ADT filtering plot showing cells to keep /remove using either RNA and/or ADT filters 
- [ ] Expression of highly variable ADTs as a ridge plot 
- [ ] Expression of highly variable ADTs as a UMAP

Another thought I had was to create a modified version of the `scpca-nf` overview to show the inputs/ outputs with CITE-seq or multiplexed data. We would add additional input files, like the barcode tables, and include additional steps such as genetic demultiplexing or ADT filtering and normalization. 

We do want to mention that genetic demultiplexing is part of the workflow. Currently we don't have any figures that show demultiplexing results in the QC report, but we could create some from an example library for this paper. We could show the demultiplexing results on a UMAP for each of the different methods that we use in a supplemental figure.

allyhawkins:
After some discussion, we plan on showing things related to processing additional modalities as a supplemental figure to Fig. 2. Fig.2 will focus on describing the pipeline and Fig. S2 will show modified versions of the pipeline with different modalities (CITE, multiplexed, and bulk). This will include: 

- A modified version of the workflow with CITE-seq
- mini versions of the plots in the QC report related to ADT/CITE-seq data
- (?) Illustrations of download structure of CITE-seq with AnnData, since it differs from SCEs
- Overview of the multiplexed workflow 
- Illustration of multiplexed download 
- Bulk workflow 

One thing that isn't reflected here is spatial. We may consider creating a workflow diagram for this as well and including it. We also may need to break this up into multiple supplemental figures, maybe one per modality depending on space.

allyhawkins:
Closed by #36

PR #36, opened 2024-01-19
FigS2B: Mini ADT plots

allyhawkins:
Related to #8 

This PR adds in a script to create the mini versions of the ADT plots for the supplemental figure covering additional modalities. Similar to how we have mini and simplified versions of the plots in the main QC report, this creates similar looking figures but for the ADT plots. 

- I added some lines to the `sync-data-files.R` script to download the data files for `SCPCS000216` for creation of these plots.
- I used the same theme settings that we used to create the mini QC plots. 
- I removed extra text or lines that may be hard to read or confusing for these smaller versions. Specifically, I removed the N=X in the filtered plot showing the number of cells that are included. I also removed the dotted line, similar to how we did that in the simplified main QC reports. 
- I removed the legend for the UMAPs, again following suit with what we did before. 
- The code for the actual plots is pretty much copied from the `cite_qc.rmd` file in `scpca-nf` with some slight alterations. 

Again, I think it's going to be hard to fully decide on if we like the size until we put things in illustrator with the overview diagram, but I think this is a good start to at least get the look of the plots done.

Tagging @jaclyn-taroni for any aesthetic comments and then requesting @sjspielman for code review.


scripts/FigS2B_adt-plots.R
```diff
@@ -0,0 +1,143 @@
...
+  # make sure Keep is first 
+  filter_summary = forcats::fct_relevel(filter_summary, "Keep"),
+  keep_column = ifelse(
+    stringr::str_starts(filter_summary, "Keep"),
+    "Keep",
+    "Remove"
+  ))
+
+# faceted plot to show cells belonging to each filter group
+filtered_plot <- ggplot(filtered_coldata_df, aes(x = detected, y = subsets_mito_percent, color = keep_column)) +
```
sjspielman, line 65: can you space this out onto multiple lines for readability?

scripts/FigS2B_adt-plots.R
```diff
@@ -0,0 +1,143 @@
...
+  facet_wrap(vars(ADT), nrow = 2) +
+  labs(x = "Log-normalized ADT expression",
+       y = "Density") +
+  theme(legend.position = "none")
+
+# ADT expression UMAP ----------------------------------------------------------
+
+# create data frame of UMAPs and expression
+umap_df <- scuttle::makePerCellDF(processed_sce) |>
+  tibble::rownames_to_column("barcode") |>
```
sjspielman, line 115: I feel like we might already actually have a column `barcodes` in the colData so you might be able to skip this, but definitely check if I'm right.

scripts/FigS2B_adt-plots.R
```diff
@@ -0,0 +1,143 @@
...
+      aspect.ratio = 1
+    )
+)
+
+set.seed(2024)
+
+
+# Set up -----------------------------------------------------------------------
+
+# define file paths and read in sce objects
```
sjspielman, line 28: I might define `sample_id` and `library_id` first and then use below? But for a one-off, doesn't seem super necessary. Up to you.

scripts/FigS2B_adt-plots.R
```diff
@@ -0,0 +1,143 @@
...
+
+# Get the top 4
+top_adts <- adt_var[order(adt_var, decreasing = TRUE)[1:4]] |>
+  names()
+
+# ADT expression density -------------------------------------------------------
+
+# grab expression for top ADTs from counts
+var_adt_exp_df <- logcounts(altExp(processed_sce))[top_adts, ] |>
+  # as.matrix needs to come _first_
```
sjspielman, line 92: I too have learned this the hard way.

sjspielman (commented):
Not a full review! Just some initial tiny thoughts from scrolling through! Actual review next week after aesthetic comments are in.

scripts/FigS2B_adt-plots.R
allyhawkins: 😂 I believe this is your original comment since I did copy some of this code

sjspielman: Past me did a good job!!

allyhawkins:
@sjspielman I responded to your initial comments. We do already have a `barcodes` column so I adjusted throughout to use that existing column and consistently use `barcodes` not `barcode`.

scripts/figure_setup/sync-data-files.R
```diff
@@ -3,14 +3,22 @@
...
 
 sync_call <- glue::glue("aws s3 cp '{qc_s3_files}' '{qc_local_files}' --exclude '*' --include '*.rds' --recursive")
 system(sync_call)
 
+# sync results for SCPCS000216 (for ADT plots)
+adt_s3_files <- "s3://nextflow-ccdl-results/scpca-prod/results/SCPCP000007/SCPCS000216"
+adt_local_files <- here::here("s3_files", "SCPCS000216")
+fs::dir_create(adt_local_files)
+
+sync_call <- glue::glue("op run -- aws s3 cp '{adt_s3_files}' '{adt_local_files}' --exclude '*' --include '*.rds' --recursive")
```
jaclyn-taroni, line 19: I think you've broken with convention here:

```suggestion
sync_call <- glue::glue("aws s3 cp '{adt_s3_files}' '{adt_local_files}' --exclude '*' --include '*.rds' --recursive")
```

figures/pngs/FigS2B_adt-plots.png
jaclyn-taroni, line 1: LGTM. Just flagging that I think we will need a very good legend for the filtering plot here!

scripts/FigS2B_adt-plots.R
```diff
@@ -0,0 +1,144 @@
...
+
+renv::load()
+
+library(ggplot2)
+library(SingleCellExperiment)
+
+theme_set(
+  theme_classic() +
+    theme(
+      strip.background = element_rect(fill = "transparent"),
```
sjspielman, line 12: The strip borders look a little thick in the final image, this seems good when I run on my end.

```suggestion
      strip.background = element_rect(fill = "transparent", linewidth = 0.5 ),
```

scripts/FigS2B_adt-plots.R
```diff
@@ -0,0 +1,144 @@
...
+      axis.line = element_blank(),
+      axis.ticks = element_blank(),
+      axis.text = element_blank(),
+      # add a square around each of the plots
+      panel.background = element_rect(colour = "black", linewidth=0.5),
+      aspect.ratio = 1
+    )
+)
+
+set.seed(2024)
```
sjspielman, line 23: I don't think we need a seed in this script

scripts/FigS2B_adt-plots.R
```diff
@@ -0,0 +1,144 @@
...
+  coord_fixed() +
+  theme(
+    legend.position = "none"
+  )
+
+# Combine and export -----------------------------------------------------------
+
+combined_plot <- patchwork::wrap_plots(list(filtered_plot, adt_density_plot, adt_umap_plot))
+
+ggsave(output_plot_file, combined_plot)
```
sjspielman, line 144: When I run the script, the text doesn't fit - and it's because my exported figure size is different from yours. We'll need height & width arguments here to ensure everything fits, as well as associated reproducibility 

FigS2B_adt-plots

sjspielman (commented):
Mostly looks good, just a few small-ish comments!

While you're here, can you also add a line in the top comments in `generate-figure-tables.sh` to run `Rscript/scripts/sync-reference-files.R`? Thanks!

allyhawkins:
I updated the linewidth, set the figure size, and added a comment about running the set up script for reference files. This should be ready for another look @sjspielman.

figures/pngs/FigS2B_adt-plots.png
sjspielman, line 1: Actually, I think maybe we might want to make the line around the density plot a bit thinner too?

sjspielman (approved):
LGTM! The only thing I might tweak is making the density plot outline a bit thinner, use your judgement! But, I don't think I need to see again unless you want me to look so going ahead and approving now.

PR #38, opened 2024-01-22
Fig 4B: SingleR vs. CellAssign heatmap

### LINKED ISSUE #7: Figure 4: Summary of cell type annotation
This figure should show an overview of the types of cell type annotation included with each sample and the associated diagnostic information. Here we should at least show plots of both SingleR and CellAssign, but I'm on the fence about including submitter annotations. We only have one project that has submitter annotations currently, so that might be better as a supplemental figure? 

Things we should include in the main figure at least are: 

- [ ] UMAP showing cell type annotations with SingleR
- [ ] UMAP showing cell type annotations with CellAssign
- [ ] Comparison of cell type annotations between both methods 
- [ ] Comparison of cell type annotations to clusters 

We probably also want to include a supplemental figure with the diagnostic plots for delta median and the CellAssign probability.

allyhawkins:
Closes #7 

This PR adds in the plot comparing annotations from SingleR vs. CellAssign for a representative library. 

- I chose to use `SCPCS000001` since we use that for the mini QC report plots and I like that the neurons and macrophage/ microglia match up. It also shows the T cells in CellAssign spread across multiple classes of T cells in SingleR. 
- I added a script here that reads in the data for this sample and creates the report pretty much in the exact same way we create the heatmap in the report. The only difference here is that I'm hard coding the label sizes and the height/width since we don't need to predict that here. 
- My understanding is that this is the only plot besides the cell type workflow overview that will be in the main figure. We will include other figures (such as comparison to submitter annotations and diagnostic plots) in the supplemental report. Because of that, I took some of the functions that we use in creating the cell type report that will be used to make more than just this plot into a separate script that gets sourced in. 
- When working on this, I noticed that we were still plotting the `Unclassified cells`. These are the cells that were not classified because they were filtered out in previous runs. We have a warning in the report that when using previously calculated CellAssign and SingleR results there might be slight differences between the cells that are present in the processed object and the annotations, so we don't show those cells in the plots. Here, I found that they were actually in the heatmaps still, so I removed them from this heatmap. I'll file a PR to `scpca-nf` with the same fix to those functions to remove `Unclassified cells` from the plots. 
- The only thing I'm iffy on is the legend placement. I tried the top and the right side, and I _think_ I like the right side better. I'm curious if others have thoughts. See https://github.com/AlexsLemonade/scpca-paper-figures/commit/8065272746f0c9b481b1a2be814f9af1edfa3803 to look at what the plot looks like with the legend on the top instead. 

Tagging @jaclyn-taroni for any aesthetic comments and @sjspielman for code review.


scripts/utils/celltype-plot-helper-functions.R
```diff
@@ -0,0 +1,140 @@
+# Helper functions for creating cell type plots 
+# These functions are taken directly from `scpca-nf/templates/celltypes_qc.rmd`,
+# `scpca-nf/templates/celltypes_supplemental_report.rmd`, and `scpca-nf/templates/utils/celltype_functions.Rmd`
```
sjspielman, line 3: I might add a permalink too. 

```suggestion
# These functions are taken directly from `scpca-nf/templates/celltypes_supplemental_report.rmd` and `scpca-nf/templates/utils/celltype_functions.rmd`
# https://github.com/AlexsLemonade/scpca-nf/tree/f8905ef8601f017219e0f8e315a22a37fded75b8/templates/qc_report
```

scripts/utils/celltype-plot-helper-functions.R
```diff
@@ -0,0 +1,140 @@
...
+
+library(SingleCellExperiment)
+
+# Set up cell type dataframe ---------------------------------------------------
+
+#' Create `celltype_df` data frame for use in cell type QC reports
```
sjspielman, line 9: I'd get rid of this function (unless we need it for other plots in the paper?) and just include this in the Fig4B script directly - 

Actually, I'd just get rid of this function - unless we'll need it for another plot in the paper? - and just put this in the Fig4B script directly (with proper spacing)

```
celltype_df <- processed_sce |>
  scuttle::makePerCellDF(use.dimred = "UMAP") |>
  dplyr::select(
     UMAP1 = UMAP.1, 
     UMAP2 = UMAP.2
     singler_celltype_annotation, 
     cellassign_celltype_annotation
    ) |>
     dplyr::filter(
       singler_celltype_annotation != "Unclassified cell",
        cellassign_celltype_annotation != "Unclassified cell"
      ) |> 
      prepare_automated_annotation_values(singler_celltype_annotation)
      prepare_automated_annotation_values(cellassign_celltype_annotation)
```

scripts/utils/celltype-plot-helper-functions.R
sjspielman, line 1: Will these functions be used elsewhere? If not, I might just keep them at the top of the Fig4B script without having to source another file.

sjspielman (commented):
Overall looks good! My main comment is whether we actually need to be sourcing a separate function file. IMO this is only necessary if we're going to use those functions again for other plots. 

Also, I don't think it makes sense to keep the full function for creating `celltype_df` again unless we might need more of its functionality for other figure papers. If this is a one-off, we can just make this df directly in the 4B script.

allyhawkins:
Just noting that, yes, we will need these functions to create the diagnostic plots and additional heatmaps that we plan on including in the supplemental. If we end up not using all of them, then I think we can put them into the script, but for now let's keep them separate. I noted which figures will be creating using those functions and left a TODO to add any additional figures that get added.

sjspielman:
>  I noted which figures will be creating using those functions and left a TODO to add any additional figures that get added.
👍 

All that said, we can probably get away with removing the "submitter" code from those functions, but I think it's best to just leave as is. We are clear where the functions come from in the first place!

sjspielman:
Also, meant to say re-legend location - I think both top and right are fine, but it depends mostly on how Fig 4 gets compiled over all. I say leave it on the right and if, when making the full figure, it becomes clear that top would be better for the final Fig 4 as a whole, we can move it then.

allyhawkins:
> All that said, we can probably get away with removing the "submitter" code from those functions, but I think it's best to just leave as is. We are clear where the functions come from in the first place!

I left the submitter code in because we're going to have a heatmap looking at submitter annotations vs. SingleR and CellAssign, so I didn't want to forget to add it later. 

> Also, meant to say re-legend location - I think both top and right are fine, but it depends mostly on how Fig 4 gets compiled over all. I say leave it on the right and if, when making the full figure, it becomes clear that top would be better for the final Fig 4 as a whole, we can move it then.

This sounds good to me.

sjspielman (approved):
🚀 🎨

Issue #37, opened 2024-01-22
Supplemental cell type figures


allyhawkins:
Breaking this out of #7, we will want a supplemental figure for cell type annotation that includes: 

- Example diagnostic plots for CellAssign and SingleR
- Comparison of submitter annotations to SingleR and CellAssign
- Some TBD information on how we decided to use SingleR and CellAssign and the references.

PR #39, opened 2024-01-23
Fig S4A-B: Diagnostic plots for cell type annotation

allyhawkins:
Addresses #37 

This PR adds the first set of supplemental cell type figures, the example diagnostic plots for SingleR and CellAssign. 

- I added a single script that produces two individual files, one for the SingleR plot and one for the CellAssign plot. 
- Again, I took most of the code directly from the supplemental cell type report. There's not a ton of variation in what we already had there. 
- I played around with the idea of filtering out cell types that have low numbers of cells, because honestly these figures are really big with all the cell types. So I tried filtering out any cell types that have < 10 cells and personally I think it makes these plots easier to interpret and read in the context of a paper. But these also will be in the supplemental instead of main, so I could see why we might want to include all cell types. 

For reference, this commit contains the full plot: https://github.com/AlexsLemonade/scpca-paper-figures/commit/28c30606371d0826ae3351e67251883289c16c81

This commit contains the filtered plots: https://github.com/AlexsLemonade/scpca-paper-figures/commit/40e49f6114936fb825c3e8d1a84fefdacdb11046

@jaclyn-taroni do you have a preferred plot?


scripts/FigS4A-B_celltype-diagnostic-plots.R
```diff
@@ -0,0 +1,199 @@
...
+  # if pruned.labels are NA ==> low confidence
+  # so, negate for this variable:
+  confident = !is.na(metadata(processed_sce)$singler_result$pruned.labels)
+) |>
+  dplyr::mutate(
+    confident = ifelse(confident, "High-quality", "Low-quality")
+  )
+
+
+# we use inner_join b/c the above tibble does NOT contain "Unknown cell type", which
```
sjspielman, line 58:
```suggestion
# we use inner_join b/c celltype_df does NOT contain "Unknown cell type", which
```

scripts/FigS4A-B_celltype-diagnostic-plots.R
```diff
@@ -0,0 +1,199 @@
+# This script is used to generate example diagnostic plots for CellAssign and SingleR 
+
```
sjspielman, line 2: Since much of this code is from the QC, let's state that and link - 

https://github.com/AlexsLemonade/scpca-nf/blob/cf1bdae3a711a37d67aa18ca1b8e26314c8e43f7/templates/qc_report/celltypes_supplemental_report.rmd

scripts/FigS4A-B_celltype-diagnostic-plots.R
```diff
@@ -0,0 +1,199 @@
...
+    y = annotation_wrapped,
+    shape = confident,
+    alpha = confident
+  ) +
+  ggforce::geom_sina(
+    size = 0.8,
+    color = "black", # will get applied to all confident points and non-confident outline
+    fill = "white", # will apply to non-confident fill only
+    position = position_dodge(width = 0.05) # Keep both types of points mostly in line
+  ) +
```
sjspielman, line 99: If we go with the filtered version, maybe we can tweak something here. See how the low confidence points have a wider y-range? They don't look quite along the line anymore. I think the `maxwidth` or `bandwidth` args would be good to try out - https://ggforce.data-imaginist.com/reference/geom_sina.html

sjspielman (commented):
> I played around with the idea of filtering out cell types that have low numbers of cells, because honestly these figures are really big with all the cell types. So I tried filtering out any cell types that have < 10 cells and personally I think it makes these plots easier to interpret and read in the context of a paper. But these also will be in the supplemental instead of main, so I could see why we might want to include all cell types.

Yeah, I see both arguments here... Let's see what Jackie thinks!

jaclyn-taroni:
Hmm. I'm partial to the full plot without knowing the _context_ (i.e., results text around it). The filtered one seems almost overly optimistic to me.

allyhawkins:
I updated this to show all cell types and remove the filtering I had previously. I also added the link to the supplemental report at the top per @sjspielman's suggestion.

scripts/FigS4A-B_celltype-diagnostic-plots.R
```diff
@@ -0,0 +1,196 @@
...
+    data = delta_median_confident_df,
+    color = "red",
+    geom = "point",
+    fun = "median",
+    shape = 18,
+    size = 2.25,
+    alpha = 0.9
+  ) +
+  guides(
+    alpha = FALSE,
```
sjspielman, line 120: ggplot2 is giving me a warning here (probably a thing we missed in the qc report b/c warnings may have been off for the chunk?)

```
Warning message:
The `<scale>` argument of `guides()` cannot be `FALSE`. Use "none" instead as
of ggplot2 3.3.4. 
```

sjspielman, line 120:
```suggestion
    alpha = "none",
```

sjspielman (approved):
🚀 , just one argument to update!

PR #49, opened 2024-02-02
Compile Figure 1 in illustrator

### LINKED ISSUE #15: Fig 1C: Example project card
As part of Figure 1, we want to include an example of a project card from the portal. This way we can point out any additional information that is available for each project, like external accessions, publications, usage restrictions, etc. We might want to just use the figma files to create this figure.

### LINKED ISSUE #42: Combine panels for Figure 1
We will need to combine the panels for Figure 1 into a single file. The file should be letter size - 8.5 x 11.

allyhawkins:
Closes #42 
Closes #15 

Here I combined the panels for Figure 1 into a full figure on illustrator. I took a look at the [Author guidelines for figures for Cell Genomics](https://www.cell.com/figure-guidelines) as some guidance here. 

- I updated the scripts to save the files as pdfs in addition to pngs. Do we want to keep both or should we just convert everything to pdf and remove the pngs? 
- I put all of the panels into the figure and then adjusted all the fonts so that they are Arial size 6 or size 8. Axis labels and the numbers for each bar in the sample plot are all size 6. The titles of the facets, axis titles, and legends are size 8. 
- I adjusted the linewidth for the facet boxes since they were closer to 2 to be 1. 
- Here I also added in the project card example. We made a simplified version of one of the existing project cards that has almost all additional items so that we can mention them in the text. I did not adjust the font size or style here since it comes from the Portal. Do we like this? We chose this project because it had both publications and external accessions. 
- I included both the ai file and the pdf.


jaclyn-taroni:
General comment – tagging @sjspielman too – I'd remove saving the panels as PNGs unless we also need PNGs (I vaguely recall some conversation about too many points). But I'd **_add_** saving the assembled figures as PNGs. That's likely the format that's going to get used for writing using Manubot – the PDFs are what we'll use for submission – to say nothing of making it easier to review the figures via the GitHub interface.

It's probably too late for this conversation, but I'd put the AI files somewhere centralized on Google Drive instead. I'd expect them to be big, and the version control probably doesn't help us over and above the ability to roll things back up to 30 days in Google Drive. (We could also keep certain versions forever as I understand it: https://support.google.com/drive/answer/2409045?hl=en&co=GENIE.Platform%3DDesktop#zippy=%2Cdownload-recent-versions)

figures/illustrator_files/figure_1.pdf
jaclyn-taroni, line 1: I would not expect a compiled PDF to live in `figures/illustrator_files`. I might expect them to live in `figures/compiled_figures/` with the understanding that all the other `figures` subdirectories contain panels.

figures/illustrator_files/figure_2.ai
jaclyn-taroni, line 1: I don't think this is supposed to be in this PR?

figures/illustrator_files/figure_1.pdf
jaclyn-taroni: Something's wrong with the warning symbol in panel C. I also find the drop shadow for the panel to be too light to be effective.

jaclyn-taroni: This might be an edit at the panel generation step. I can't tell what the 1(?) adrenocortical carcinoma sample is that's not an initial diagnosis. Progressive? Making the lines in the pattern itself thinner might help with this, and I think it might also benefit the Brain and CNS panel.

jaclyn-taroni (commented):
The overall layout and font sizes look good to me. I found some issues with Panel C and the progressive pattern in Panel A.

allyhawkins:
I made the following organizational changes that once approved we can apply to other figures. I'll work on the aesthetic updates next, but just wanted to get this part figured out. 

- Removed the `png` files for individual panels where we don't need them. 
- Moved the AI files to a folder on our drive, [`ScPCA/Manuscript figures`](https://drive.google.com/drive/folders/1kLtyaLZbVP4KUT6UijJm7xCGVqvKJhZE?usp=drive_link). 
- Saved the pdf and png versions of the compiled panels to `compiled_figures/pngs` and `compiled_figures/pdfs`.

jaclyn-taroni:
@allyhawkins 

* The updated linewidth is much better, thanks. Naively, I might expect that to be a code change, too, but maybe that's too difficult.
* I think you do need something to indicate that the last panel is a card of some kind – try adding back the drop shadow and darkening it

allyhawkins:
> The updated linewidth is much better, thanks. Naively, I might expect that to be a code change, too, but maybe that's too difficult.

The only way to adjust the actual pattern linewidth is to adjust spacing and density. We are already using almost the smallest version of those and they don't just adjust linewidth so I thought it would be easier to just do it in illustrator. 

> I think you do need something to indicate that the last panel is a card of some kind – try adding back the drop shadow and darkening it

I added it back and darkened it, how does it look now @jaclyn-taroni?

jaclyn-taroni (approved):
👍🏻

PR #51, opened 2024-02-02
Create compiled figure 2

### LINKED ISSUE #43: Combine panels for Figure 2
We will need to combine the panels for Figure 2 into a single file. The file should be letter size - 8.5 x 11.

allyhawkins:
Closes #43 
Stacking on #49 

This PR adds in a compiled version of Figure 2 as an AI file and PDF file. 

- I updated the script for creating Figure 2B to save as a pdf. 
- I resized the overview figure and made sure all the labels for the icons were size 6 and the bigger labels were size 8 (the SCE output and AnnData output labels). 
- I did not edit any of the icons in the overview or the text inside the icons. I can do that if we think that's necessary. 
- For the mini QC plots, all the axes titles are size 8 and the facet labels are size 6. I also adjusted the facet boxes so the linewidth is 1 instead of 2. Are we okay with the size of these smaller panels? I could make them bigger if we want them to take up the whole page, but I don't think that's necessary.


allyhawkins:
> I assume this was supposed to go to me for review.

Yes that's correct.

allyhawkins:
@jaclyn-taroni I made the same organizational changes here as mentioned in https://github.com/AlexsLemonade/scpca-paper-figures/pull/49#issuecomment-1938884084. 

I also updated the font size for the overview figure to be as big as I could get it without the text being overcrowded. I made the QC plots take up more space and thus increased the font size.

jaclyn-taroni (approved):
Very minor comment, and I don't need to see it again.

Can you make sure `alevin-fry` and `Read into R` are centered over their arrows? Thanks!

PR #52, opened 2024-02-06
Compiled Figure 4

### LINKED ISSUE #45: Combine panels for Figure 4
We will need to combine the panels for Figure 4 into a single file. The file should be letter size - 8.5 x 11.

sjspielman:
Closes #45 

This PR adds the compiled version of Figure 4. 

- I updated `scripts/Fig4B_singler-cellassign-heatmap.R` to export both PNG and PDF, and fixed a few things along the way:
  - We were missing a `grid::` somewhere
  - I had to add the heatmap argument `column_names_max_height = unit(8, "cm")` (upped from the default 6 cm) to ensure the column labels fully fit in the exported PDF
  - Turns out `ComplexHeatmap` and dependencies never made it into `renv.lock`, so I snapshotted them in.

This is being filed as a Draft since certain font size/family decisions are still in flux, pending reviews in #49 & #51. Once this has been formalized, I will update here accordingly and request review.


sjspielman:
I went ahead and updating the sizing (waiting probably not too necessary); marking as ready for review now.

figures/illustrator_files/figure_4.pdf
jaclyn-taroni, line 1: Similar comment to Figure 3 - fonts are too small and we can try reducing the whitespace to address it. https://github.com/AlexsLemonade/scpca-paper-figures/pull/59#discussion_r1486342443

sjspielman:
I've updated here with larger fonts, and the right file types (pdf/png) in the right spots without any files we don't need. How's it looking?

figures/compiled_figures/pngs/figure_4.png
jaclyn-taroni, line 1: This isn't a blocker for merging; it is more something to keep in mind as we write text (cc:@allyhawkins) – panel A almost implies that the cell type annotation is only for the SCE objects. I know that we convert to AnnData after the fact, but we should make sure our audience knows, too, when they get to this figure.

jaclyn-taroni (approved):
👍🏻

figures/compiled_figures/pngs/figure_4.png
sjspielman: Something for figure legend, too.

PR #53, opened 2024-02-06
Compile supplemental Figure 1

### LINKED ISSUE #46: Combine panels for Supplemental Figure 1
We will need to combine the panels for Supplemental Figure 1 into a single file. The file should be letter size - 8.5 x 11.

allyhawkins:
Closes #46 

This adds the compiled figure for supplemental figure 1, the benchmarking figure. 

- As with the other PRs, I updated the script to export pdfs alongside the pngs for each of the panels. 
- I updated text so that everything is arial. Size 8 for axis titles and legends, size 6 for axis labels and facet labels. 
- Originally we were going to have the order be time/memory plots, UMI totals, genes detected, and then gene expression as panel D. However, the UMI and genes detected plot are similar to each other so having them on the same row looked better to me. So I chose to have panel B be the gene expression correlation and then panel C/D be UMI and genes detected. It just means we should talk about gene expression first in the text, which I think is fine.


allyhawkins:
@jaclyn-taroni I've updated the organization of the files here to be the same as in https://github.com/AlexsLemonade/scpca-paper-figures/pull/49#issuecomment-1938884084.

jaclyn-taroni (approved):
👍🏻

PR #54, opened 2024-02-06
Update sample download folder image

sjspielman:
Just swapping in the updated version of this figure with the correct `.tsv` extension.


sjspielman:
We're still on the fence for whether we want to include this image, but if we do, this is the one to include!

allyhawkins (approved):
🐱 🐱

Issue #47, opened 2024-01-24
Combine panels for supplemental Fig 2/3


allyhawkins:
We will need to combine the panels for Supplemental Figure 2 and 3 (additional modalities figures) into a single file. The file should be letter size - 8.5 x 11.

allyhawkins:
Fig S2 panels: 
A - CITE-seq workflow 
B - mini versions of the plots in the QC report related to ADT/CITE-seq data
C - multiplexed download 

Fig S3 panels: 
A - bulk workflow
B - spatial workflow 

At one point we discussed including illustrations of the associated downloads, but I think we might be okay without them.

sjspielman:
> At one point we discussed including illustrations of the associated downloads, but I think we might be okay without them.

Is this referring to what you listed as Fig S2C? I don't see a multiplex download image in the repo

allyhawkins:
> > At one point we discussed including illustrations of the associated downloads, but I think we might be okay without them.
> 
> Is this referring to what you listed as Fig S2C? I don't see a multiplex download image in the repo

Sorry, typo! I meant the multiplexed workflow.

sjspielman:
@allyhawkins let me get your opinion on this - 

For FigS2, the ggplot panel is a bit domineering, but it's basically just what it is given the `aspect.ratio = 1` which we need for the UMAP panels. So I wondered if we might want to reorder be either: a/b are workflows and c is the plots, or a is the plots and b/c are workflows. _Conceptual_ screenshots below. Any thoughts?
 
<img src="https://github.com/AlexsLemonade/scpca-paper-figures/assets/4701111/9bed3322-a4f2-49d3-9ebc-635e7333808e" width="200"/>

<img src="https://github.com/AlexsLemonade/scpca-paper-figures/assets/4701111/b0a8d552-72e3-4a98-8a9d-7dcc531bbffe" width="180"/>

<img src="https://github.com/AlexsLemonade/scpca-paper-figures/assets/4701111/07625bb4-cdd9-495f-887b-61c1dd479260" width="180"/>

sjspielman:
Ooof also, saved as an AI file it's about 47 MB - probably because of all the _points_ in the plots... Since this is a supplemental figure, maybe it's fine to just place a TIFF or PNG export of that ggplot panel in to reduce file size? We did this for some OpenPBTA supp figures that got too big.

allyhawkins:
[...]

PR #55, opened 2024-02-07
Compile figure S2

sjspielman:
Part of #47 
Edit - Stacked on #52 

This PR includes the compiled version of Figure S2. Here, I did do all text resizing to match. Everything is 6 pt Arial, except for the plot _axis titles_ in Panel C which are 8 pt Arial. Notably, I kept Panel C as a png, since it is _huge_ when exported as a PDF due to all the many scatterplot points. Then the compiled figure is even bigger (approaching 50 MB). The resolution with tiff wasn't much different from png either, so I kept it as png. Then in Illustrator, I replaced labels with little text boxes in Arial font at the right size.


jaclyn-taroni:
@sjspielman can you update this to reflect our decisions earlier in the week, e.g., remove the `.ai` file and make the fonts bigger since we seem to have room to do that?

sjspielman:
This has been updated with larger fonts. As before, Panel C is a PNG that is placed into the illustrator; otherwise the figure file size gets too big (approaching 40 MB). Since this is a supplemental figure, that choice seems not so bad!

figures/compiled_figures/pdfs/figure_S2.pdf
jaclyn-taroni, line 1: `figure_s2.pdf` (also applies to PNG) would better conform with the convention in `main`.

jaclyn-taroni (approved):
👍🏻  Minor comment: please update the file names to be more consistent with what is in `main`. I don't need to take another look.

PR #57, opened 2024-02-07
Compiled figure S3

sjspielman:
Rest of #47 
Here is Fig S3, but it has _lots_ of whitespace if we want everything at 8.5x11!

Here's also a screenshot of what it _might_ look like if we wanted to, for example, fill some whitespace with download images, but as @allyhawkins pointed out in https://github.com/AlexsLemonade/scpca-paper-figures/issues/47#issuecomment-1932970232, if we add the download images here we'd probably want to do it for all figures with workflows, which might not be something we actually want.


sjspielman:
This is now updated with a larger font, with only panels A-B and at paper size 8.5x11 as per the original issue. There is inevitably a lot of whitespace with this paper size, so if we have other ideas about filling this whitespace (or accepting it as it is what it is?) let me know!

sjspielman:
Argh I borked this branch because I thought I had it stacked so I merged something into it, but that was wrong! Going to close out this PR and open a fresh one.

PR #58, opened 2024-02-09
Script for creating merged UMAPs for figure 3

### LINKED ISSUE #9: Figure 3: Inclusion of merged objects and downloading projects
By the time we submit this paper, we hope to have made merged objects available on the portal. Along with that we will have a report that is included with the merged objects. The content of this report are TBD, but I assume we will want to show some of that content in a figure. 
I'm filing this as a placeholder right now, because I think once we discuss that report we will have a better idea of what we might want to show here.

allyhawkins:
Closes #9 

This PR adds in an example of the UMAPs that we are including for merged UMAPs. I know that we haven't actually produced any final merged objects, but we still plan to show these UMAPs. I used one of the projects that we've been testing with and only am showing 4 libraries, which I think is plenty. 

- I updated the script to sync data files to include syncing the merged object for this project. Note that right now this is using the object that lives in `s3://nextflow-ccdl-results/scpca/processed` rather than `scpca-prod`, but I think that's okay. If we wanted to use the prod version, we would just need to switch the directory that's listed in the script. 
- In the script to create the UMAPs I picked 4 libraries that we want to show and am filtering to those 4 libraries only before plotting. This should control for any future changes in the merged object if we were to include more libraries or use the prod version with all libraries. 
- As with a lot of the other plots, I took most of the code for the plot itself from the merged report template in `scpca-nf`. 

Tagging @jaclyn-taroni for any aesthetic comments and requesting @sjspielman for review.


scripts/Fig3D_merged-umaps.R
```diff
@@ -0,0 +1,90 @@
...
+      axis.line = element_blank(),
+      axis.ticks = element_blank(),
+      axis.text = element_blank(),
+      # add a square around each of the plots
+      panel.background = element_rect(colour = "black", linewidth=0.5),
+      aspect.ratio = 1
+    )
+)
+
+set.seed(2024)
```
sjspielman, line 26: not needed unless we are _calculating_ the UMAP, which we're not!

scripts/Fig3D_merged-umaps.R
```diff
@@ -0,0 +1,90 @@
...
+png_dir <- here::here("figures", "pngs") 
+output_png_file <- file.path(png_dir, "Fig3D_merged-umaps.png")
+
+pdf_dir <- here::here("figures", "pdfs") 
+output_pdf_file <- file.path(pdf_dir, "Fig3D_merged-umaps.pdf")
+
+
+# Prep for plotting ------------------------------------------------------------
+
+libraries_to_plot <- c("SCPCL000050",
```
sjspielman, line 49: I wonder if this should be in the `setup` portion of the script, since I imagine this might change later.

scripts/README.md
```diff
@@ -60,11 +60,14 @@ Before running this script, you must run `figure_setup/sync-data-files.R` and `s
 7. `FigS2B_adt-plots.R`: This script is used to generate supplemental Figure 2B, which includes simplified and miniature versions of the plots in the ADT section of the main QC report included with each sample download.
 Before running this script, you must run `figure_setup/sync-data-files.R`.
 
-8. `Fig4B_singler-cellassign-heatmap.R`: This script is used to generate Figure 4B, which includes an example of the heatmap comparing cell type annotations from `SingleR` to `CellAssign`.
+8. `Fig3D_merged-umaps.R`: This script is used to generate Figure 3D, which includes an example of the UMAPs showin the merged report.
```
sjspielman, line 63:
```suggestion
8. `Fig3D_merged-umaps.R`: This script is used to generate Figure 3D, which includes an example of the UMAPs shown in the merged report.
```

scripts/figure_setup/sync-data-files.R
```diff
@@ -63,3 +63,13 @@ fs::dir_create(celltype_local_files)
 
 sync_call <- glue::glue("aws s3 cp '{celltype_s3_files}' '{celltype_local_files}' --exclude '*' --include 'SCPCL000498_processed.rds' --recursive")
 system(sync_call)
+
+# sync merged object results for SCPCP000003 -----------------------------------
```
sjspielman, line 67: I wonder if generally we should update this script to use `aws s3 sync` instead of `aws s3 cp`? It's my understanding (am I correct?) that `cp` will _always_ copy, but `sync` will only copy if the local file doesn't exist or is out of date.

sjspielman: Also, I think some of the other syntax (like `--recursive`) would change?

allyhawkins: Just wanted to note that I've been burned before by `sync`, because it doesn't always update if the file is out of date. I had this issue when benchmarking alevin-fry, which is why I've been using `cp` for all of this.

scripts/Fig3D_merged-umaps.R
allyhawkins: Sorry this was leftover from copying this set up chunk, I'll remove it.

scripts/Fig3D_merged-umaps.R
allyhawkins: I think my plan is to not change this later. I don't think it matters which libraries we use, I just want to make sure that if the merged object gets run on more libraries, we are always picking the same 4.

scripts/figure_setup/sync-data-files.R
sjspielman: Maybe workaround with `if` then? Only `cp` if the target files don't exist?

sjspielman: Though I guess then we won't get new files if things are out of date. I wonder if there are _some_ files this would help for then, if not all.

allyhawkins: I think we generally want to make sure we are copying over the up to date files. The only files that I imagine will never change, mainly because we deleted the FASTQ files, are the alevin-fry and cell ranger output. So we could add an `if` statement around those, but I also don't know if that's totally necessary.

allyhawkins: > To be fair, my comment was brought to you by, "ugh so much copying until i get to copy the file i want to test the code"

This is a valid point. We could break up that script or add options so it only copies what you need? So a flag for each of the items that need to be copied and only copy them if that flag is used?

sjspielman: Maybe... The best idea I can think of for flags is a figure name? So like, `--fig_3` will copy files needed for that figure only, and `--all` would do everything. But, this also seems like a not great use of time considering the code works, and I am capable of doing some quick script changes when I only want _some_ files! Plus, this is probably only relevant to running code during development.

Maybe the right approach is to have a comment up top: This will always copy all files regardless of whether they are on your system already, and justt move on.

jashapiro: How often are these files changing? This is just for the paper, right? I'd probably switch to `sync` and then occasionally delete files to make sure they are up to date. Note thought that `sync` is for directories, so individual file downloads might have to be handled differently (with `--include`)

I also want to believe that the earlier problems with `sync` are in the past and we can trust it again. If we want to be more conservative, we can also use `--exact-timestamps` which will download the file even if the local copy is newer (if it was `touch`ed somehow for example). Finally, before any publication runs, you could delete the local files to force a new download.

allyhawkins: Based on this discussion, I went ahead and moved everything to be `sync` with `--exact-timestamps` rather than `cp`. I also updated the script for generating figures to include running the scripts to download the data before generating the figures. I added a step to remove `s3_files` before running the download scripts so that everything is fresh. These changes are in https://github.com/AlexsLemonade/scpca-paper-figures/pull/58/commits/a295eadec9a25f333b2392c9669a7896cd7a39d1 and https://github.com/AlexsLemonade/scpca-paper-figures/pull/58/commits/3bd8250e1b137401f2d7b374bb14fc5e15e0320b

generate-figures-tables.sh
```diff
@@ -25,8 +19,21 @@ script_dir=${BASEDIR}/scripts
...
 mkdir -p ${BASEDIR}/tables
 
+# check if s3 files directory already exists
+# if it does, delete it
+if [ -d "${BASEDIR}/s3_files" ]; then
+  rm -r ${BASEDIR}/s3_files
+fi
+mkdir -p ${BASEDIR}/s3_files
+
+# copy over any files needed from S3
```
sjspielman, line 32: since we're deleting the directory, it's all of them!
Also, since these steps take a long time, maybe let's print a status message...

```suggestion
# copy over files from S3
echo "Downloading data files..."
```

scripts/Fig3D_merged-umaps.R
```diff
@@ -0,0 +1,88 @@
...
+  theme_classic() +
+    theme(
+      #plot.margin = margin(rep(20, 4)),
+      strip.background = element_rect(fill = "transparent", linewidth = 0.5),
+      # no axis ticks or labels
+      axis.line = element_blank(),
+      axis.ticks = element_blank(),
+      axis.text = element_blank(),
+      # add a square around each of the plots
+      panel.background = element_rect(colour = "black", linewidth=0.5),
```
sjspielman, line 21:
```suggestion
      panel.background = element_rect(colour = "black", linewidth = 0.5),
```

generate-figures-tables.sh
```diff
@@ -25,8 +19,21 @@ script_dir=${BASEDIR}/scripts
...
+Rscript ${script_dir}/figure_setup/sync-data-files.R
+Rscript ${script_dir}/figure_setup/sync-metadata.R
+Rscript ${script_dir}/figure_setup/sync-reference-files.R
+
 #########################################################
 #        Generate figures in order of appearance        #
 #########################################################
```
sjspielman, line 39:
```suggestion
#########################################################

echo "Generating figures and tables..."
```

sjspielman (approved):
Updated script with `sync` works for me, everything looks good on my end!

figures/pngs/Fig3D_merged-umaps.png
jaclyn-taroni, line 1: 👍🏻

PR #59, opened 2024-02-09
Compile Figure 3

### LINKED ISSUE #44: Combine panels for Figure 3
We will need to combine the panels for Figure 3 into a single file. The file should be letter size - 8.5 x 11.

allyhawkins:
Closes #44 
Stacked on #58 

This PR adds in a compiled version of Figure 3. Originally, we had talked about showing the individual SCE object download structure, then the merged workflow, and then the merged download illustration, but that doesn't look the best on the page. I chose to show both download structures as A and B and then the merged workflow as C. 

- Since we do want to show the download illustrations here for both individual SCEs and merged SCEs, I added those pngs to the repo here. 
- For the merged workflow, I changed all the icon labels to Arial size 8. 
- For the UMAPs, I chose to use the png and then re-do each of the labels. The pdf was very large since it's 4 UMAPs that each have a merged object, so lots of little points. Here all the text is Arial size 8.


figures/illustrator_files/figure_3.pdf
jaclyn-taroni, line 1: Top-level comment: I don't think this needs as much white space, and that is a relatively easy path to more legible font sizes, in my opinion.

I would:

1. Make panels A and B larger. I assume that should make the font sizes bigger "for free."
2. Make panel C larger, mostly to facilitate bumping the font size.
3. Increase the font sizes in Panel D. I don't think the panel size overall needs to change.


Other comments:

- Some of the alignment in Panel C seems off to me (and different from what is in the SVG file) – `SCE Merged Output` and `AnnData Merged Output` in particular.

jaclyn-taroni (commented):
Overall layout looks pretty good. It's the fonts and their sizes that will get you every time.

figures/illustrator_files/figure_3.pdf
allyhawkins: @jaclyn-taroni regarding the font sizes for panel C and D, I was following the guidelines for Cell Genomics that all font sizes in figures need to be between size 6-8 point in the desired print size. So all the font for C and D are size 8. Should I not be following that and making them bigger throughout?

jaclyn-taroni: Let's discuss at pre-planning

allyhawkins:
@jaclyn-taroni I made panels A, B, and C bigger so the font sizes should be bigger. I also readjusted the alignment for the output data labels in panel C and then increased the font size for panel D. 

The last thing I did was update the organization to follow https://github.com/AlexsLemonade/scpca-paper-figures/pull/49#issuecomment-1938884084.

jaclyn-taroni (approved):
👍🏻  One thing we might want to keep track of/an eye on is if the PNGs underlying panels A and B should be saved at a higher quality in the future. My PDF to PDF conversion senses are tingling. But I don't think that needs to hold things up.

PR #61, opened 2024-02-14
Compiled figure s3

sjspielman:
Part of #47 

This PR adds the compiled figure S3 with larger font (except for the summary report font, which won't fit otherwise - any thoughts there?).

There is quite a bit of (inevitable?) whitespace since I used 8.5x11, but I could also rotate and make the figure landscape instead, but these workflows are relatively small so that seemed like it might be " too aggressive," as it were. 

If there any thoughts on how to handle whitespace, beyond just accepting that it is what it is, let me know!


jaclyn-taroni:
Hm, what about a cropped version as a PNG? If we're thinking about the PNG versions as mostly relevant to Manubot...

sjspielman:
> Hm, what about a cropped version as a PNG?

👍

jaclyn-taroni (approved):
👍🏻

PR #64, opened 2024-02-20
Switch order of panels for Figure S2

### LINKED ISSUE #62: Swap order of panels in Fig. S2
In the text, we want to talk about all things related to each additional modality before moving on to the next additional modality. That means we should have what is currently panel A (CITE-seq workflow) and panel C (ADT plots) be back to back as panel A and B. 

What is currently panel B, the multiplex workflow should be C. 
See https://github.com/AlexsLemonade/ScPCA-manuscript/pull/20#discussion_r1490110021

allyhawkins:
Closes #62 

Here I switched panels B and C so that the ADT plots are panel B and the multiplex workflow is panel C. While I was here, I noticed the linewidths didn't match within the diagrams so I updated all of those to match. I also updated the font that was used around the ADT plots since it didn't appear to be Arial and seemed to just be the default from ggplot.


jaclyn-taroni (approved):
👍🏻

PR #72, opened 2024-02-27
Create tables summarizing total sample numbers for diagnosis group and disease timing

allyhawkins:
In https://github.com/AlexsLemonade/ScPCA-manuscript/pull/26, we added the total numbers of samples for each group and each disease timing. To get the total numbers, I created two tables (one for diagnosis and one for disease timing) using the data frame that we create in the script for producing Fig1. I thought we may want to be able to update those numbers in the future, so why not save the tables and the short code I used to produce them.


sjspielman (commented):
Definitely a good idea to make these tables! But, I might save them somewhere else besides `tables`, because these are not actual tables that appear in the manuscript.

What if we had a separate folder in repo like `manuscript-numbers` (is this a general enough name for other things that might be in there later?) where we export these files to? We could document in the `README` that that folder doesn't contain items that appear in the manuscript directly, but rather data that is reported in the manuscript. What do you think of an approach like that?

allyhawkins:
Good call, I moved them to `manuscript-numbers` instead and added that folder description to the readme.

PR #73, opened 2024-02-28
Fix Table S1 counts

sjspielman:
This PR fixes a bug in the counting for table S1. We were double (and then some!) counting some libraries due to carrying the diagnoses themselves forward int he counting. So, I separated out the code that creates the column for all diagnoses from the code that does the counting. I then join in this table with diagnoses, and counts seem much more sensible now for the multiplexed project!


allyhawkins (approved):
Thanks for catching this and fixing it! Looks good.

PR #71, opened 2024-02-26
Update the README

sjspielman:
This PR does some housekeeping and updates the overall `README` file to include some items it was missing. I also reorganized so that all main text figures appear before supp figures/tables.

Note that some of the changes here are reflected in open PRs #69 and #70.

PR #69, opened 2024-02-26
Compiled figure updates

sjspielman:
Closes #66 
Closes #67 

This PR updates the compiled versions of Fig 2 and S2:

#### Figure 2
I broke down panel B into panels B-G and updated the compiled figure. To improve the alignment in the compiled figure, I also separated out the final panel G, which was a little too tall in the previous compilation to easily place the panel label. This is now exported as its own PDF panel.
As such:
- `figures/pdfs/Fig2B_mini-qc-plots.pdf` is now gone
- Instead we now have `figures/pdfs/Fig2B-F_mini-qc-plots.pdf` and `figures/pdfs/Fig2G_mini-qc-hvgs.pdf`
  - Associated script updates in `scripts/Fig2B-G_qc-plots.R`

#### Figure S2
I broke down panel B into panels B-D and updated the compiled figure.

- `figures/pngs/FigS2B_adt-plots.png` is now named `figures/pngs/FigS2B-D_adt-plots.png`
  - associated script change in `scripts/FigS2B-D_adt-plots.R`

The `scripts/README.md` is also updated accordingly.

PR #74, opened 2024-03-04
Change S4C to S5

sjspielman:
Previously missed this in #69 & #71; Fig S4C is now S5, so I updated a few more spots to reflect the correct panel.


allyhawkins (approved):
🐱 🐱

PR #77, opened 2024-03-06
Updated S2 with ADT-by-cell matrix

### LINKED ISSUE #76: Consider adding an adt-by-cell matrix to Figure S2
In Figure S2, the current legend text describes that both a gene-by-cell matrix and an adt-by-cell matrix are produced and read into R: https://github.com/AlexsLemonade/ScPCA-manuscript/blob/51b1cdc8f97e738e358ed689e0b863b9faa357b4/content/100.figure-table-legends.md?plain=1#L141-L145

But, [the figure itself](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/figures/compiled_figures/pngs/figure_s2.png) does not show an adt-by-cell matrix, only the gene-by-cell matrix. 

This issue is meant to a) discuss whether we should add it to the figure, and b) if so, track adding it in. Noting that one reason to not include it is that it may make the figure too crowded, and there's already a lot going on in it.

sjspielman:
Closes #76 (boo bad branch name, the S**2** led to 72 instead of 76 😂 )

This PR adds the updated Figure S2. The only change is adding that ADT-by-cell matrix to the citeseq flowchart.


allyhawkins (commented):
This may be annoying, but can you update the actual svg for the workflow with this addition too? That way it's in sync with what's in the compiled figure.

sjspielman:
Yes, I will try!

allyhawkins (approved):
I think this is good, thank you! Just want to make sure we don't have two different versions of the panel floating around in the repo.

PR #78, opened 2024-03-07
Add table with list of references used from Panglao and associated organs

### LINKED ISSUE #75: Include supplemental table with organs used to build PanglaoDB refs
I think this a good level of detail for the text, but we might want a supplemental table of the organ sets we used.

_Originally posted by @jashapiro in https://github.com/AlexsLemonade/ScPCA-manuscript/pull/61#discussion_r1511790694_

At minimum, this table should include one row for each reference that we used and a column indicating which organs were included. 
We could also have one row per ScPCA project, and include the reference and list of organs used for that project. If we did this we would probably also want to include the diagnosis group and diagnoses in that project.

allyhawkins:
Closes #75 

Here I'm adding a supplemental table that includes the list of references created from Panglao for use with CellAssign. For each ScPCA project, I included the reference name and then the list of organs from Panglao used to create the reference. I also included a list of diagnoses for each project to hopefully allow readers to see why some of the organs are included for each project. 

To create the table, I needed to get the ref name for each project from `scpca-celltype-project-metadata.tsv`, so I added that to the `sync-metadata.R` script. I also needed the list of organs, which is stored in the `references` directory in `scpca-nf` that doesn't sync anywhere. I just added that to the repo here in the `sample-info` directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,7 @@ Summary of libraries and types of libraries found on the Portal.
 
 **Table S2**
 
-List of references used for each project on Portal with `CellAssign`, including the list of organs used to create the reference.
+List of references used for each project on the Portal with `CellAssign`, including the list of organs used to create the reference.
 
 
 ## Generating figures and tables
--- a/README.md
+++ b/README.md
@@ -76,6 +76,10 @@ Comparison between submitter provided annotations and automated annotations from
 
 Summary of libraries and types of libraries found on the Portal.
 
+**Table S2**
+
+List of references used for each project on Portal with `CellAssign`, including the list of organs used to create the reference.
+
 
 ## Generating figures and tables
 
--- a/scripts/TableS2_cellassign-ref-summary.R
+++ b/scripts/TableS2_cellassign-ref-summary.R
@@ -44,15 +44,15 @@ grouped_sample_df <- sample_df |>
   dplyr::filter(scpca_project_id %in% project_df$scpca_project_id) |> 
   dplyr::group_by(scpca_project_id) |> 
   # create a list of diagnoses for each project
-  dplyr::summarize(Diagnoses = paste(unique(diagnosis), collapse = ", ")) |> 
+  dplyr::summarize(Diagnoses = paste(sort(unique(diagnosis)), collapse = ", ")) |> 
   # add ref name
   dplyr::left_join(project_df) |> 
   # add organs
   dplyr::left_join(panglao_df, by = c("cellassign_ref_name" = "celltype_ref_name")) |> 
   dplyr::rename(
     "ScPCA Project ID" = "scpca_project_id",
-    "PanglaoDB reference name" = "cellassign_ref_name",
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -76,6 +76,10 @@ Comparison between submitter provided annotations and automated annotations from
 
 Summary of libraries and types of libraries found on the Portal.
 
+**Table S2**
+
+List of references used for each project on Portal with `CellAssign`, including the list of organs used to create the reference.
```
jashapiro, line 81:
```suggestion
List of references used for each project on the Portal with `CellAssign`, including the list of organs used to create the reference.
```

scripts/TableS2_cellassign-ref-summary.R
```diff
@@ -0,0 +1,60 @@
...
+  # create a list of diagnoses for each project
+  dplyr::summarize(Diagnoses = paste(unique(diagnosis), collapse = ", ")) |> 
+  # add ref name
+  dplyr::left_join(project_df) |> 
+  # add organs
+  dplyr::left_join(panglao_df, by = c("cellassign_ref_name" = "celltype_ref_name")) |> 
+  dplyr::rename(
+    "ScPCA Project ID" = "scpca_project_id",
+    "PanglaoDB reference name" = "cellassign_ref_name",
+    "Organs included in reference" = "organs"
```
jashapiro, line 55: I am suggesting renaming the titles here, since the names are ours, right? Or we could call this `CellAssign` reference name?

```suggestion
    "ScPCA reference name" = "cellassign_ref_name",
    "PanglaoDB organs included in reference" = "organs"
```

scripts/TableS2_cellassign-ref-summary.R
```diff
@@ -0,0 +1,60 @@
...
+  # only keep project ID and ref name for joining 
+  dplyr::select(scpca_project_id, cellassign_ref_name)
+
+grouped_sample_df <- sample_df |> 
+  dplyr::select(scpca_project_id, diagnosis) |> 
+  # remove any projects that don't have cell assign refs
+  dplyr::filter(scpca_project_id %in% project_df$scpca_project_id) |> 
+  dplyr::group_by(scpca_project_id) |> 
+  # create a list of diagnoses for each project
+  dplyr::summarize(Diagnoses = paste(unique(diagnosis), collapse = ", ")) |>
```
jashapiro, line 47: Throw in a sort?
```suggestion
  dplyr::summarize(Diagnoses = paste(sort(unique(diagnosis)), collapse = ", ")) |> 
```

jashapiro (commented):
Looks good. My only substantive suggestion is the naming of the table, but there are a couple other little things in there as well. 

If you take those, you will need to regenerate the output table though!

allyhawkins:
I liked the renaming suggestion, so I incorporated that and re-generated the table.

jashapiro (approved):
LGTM!

Issue #41, opened 2024-01-24
Supplemental figure for how we chose cell type annotation methods


allyhawkins:
As mentioned in #37, we may want to include a figure that shows some benchmarking of methods or any data on why we chose to use `SingleR` and `CellAssign`. It might also be good to include information on maybe why we chose the references that we did. I think this will be largely dependent on what we say in the text about these methods and if we feel we need something here. So I'm filing this issue to remind us that we may want to include this in the future, but I'm not entirely what the contents will be yet.

allyhawkins:
Just noting that we are going to try and include a figure like this, after some discussion in https://github.com/AlexsLemonade/ScPCA-manuscript/pull/61, we would like to include a comparison between the references tested for `SingleR`.

allyhawkins:
Okay, so for SingleR, I created a plot that shows the distribution of the delta median statistic for all cells with a given celldex reference. I grabbed three libraries for each diagnosis group, using libraries from 3 of our 4 diagnosis groups, and annotated them with SingleR with all 4 celldex references we had been using for comparisons. Then I made a plot that compares the distributions across all refs. This shows that `BlueprintEncodeData` had either the highest delta median or was equivalent to the delta median for other references. 
![FigS6_celldex-ref-comparison](https://github.com/AlexsLemonade/scpca-paper-figures/assets/54039191/8a6ab9c5-887e-445c-972a-9c9782321599)

@jaclyn-taroni @jashapiro what do we think? 

Also, I'm trying to think of a way to show a similar plot for `CellAssign`. Most of our benchmarking data came from comparing annotations with `CellAssign` to submitter annotation for just samples from the RMS project. With using different references based on tissue type, I feel like maybe we don't need anything more than the table in #75?

jaclyn-taroni:
Part of CellAssign's appeal is that it will label cells as unassigned. Is there any way to show that?

I think this plot is _fine_ conceptually, but it needs some tweaks:

- The panels will need to be bigger or points more transparent to distinguish between high-quality and low-quality effectively.
- The y-axis ticks need labels.
- Rotate the reference labels to 45 degrees – Might help with panel sizing ☝🏻

allyhawkins:
> Part of CellAssign's appeal is that it will label cells as unassigned. Is there any way to show that?

[...]

PR #79, opened 2024-03-08
Panel comparing celldex references for SingleR

allyhawkins:
Related to #41 

This PR adds the scripts and set up needed to produce the figure comparing the delta median statistic for libraries using 4 different refs from `celldex`. 

- For setup, I grabbed three libraries from three different projects (9 total) from `scpca-prod` and sync them to our `s3_files` folder. I also synced the SingleR models that we have in `scpca-references` for `celldex` so that we could run `SingleR` with all the references. 
- The script first reads in all the SCE objects and then for each object/ reference combination, runs SingleR and grabs the delta median. 
- All of the delta median scores are combined into a single dataframe that's used to generate the plot. For the plot itself, I mostly used what we use to create the delta median plot in the report, but am faceting by library and coloring the facet panels based on the `diagnosis_group`. 
- I did update this to add in the axis ticks and labels, angle the x-axis labels, and I tried to make the panels themselves bigger. Is this okay or should we adjust more? I also moved the legend to the top, which I think helps. 
FigS6_celldex-ref-comparison

Tagging @jaclyn-taroni for any aesthetic comments, and will request formal review from @jashapiro.


figures/pngs/FigS6_celldex-ref-comparison.png
jaclyn-taroni, line 1: 👍🏻

scripts/FigS6_celldex-ref-comparison.R
```diff
@@ -0,0 +1,192 @@
...
+backgrounds <- rep(c("Brain and CNS", "Leukemia", "Sarcoma"), each = 3) |>
+  purrr::map(\(x) element_rect(fill = diagnosis_colors[[x]]))
+
+# create delta median plot 
+all_delta_plot <- ggplot(all_delta_df, aes(x = celldex_reference, y = delta_median, shape = confident, alpha = confident)) +
+  ggforce::geom_sina(
+    size = 0.5,
+    color = "black", # will get applied to all confident points and non-confident outline
+    fill = "white", # will apply to non-confident fill only
+    position = position_dodge(width = 0.05) # Keep both types of points mostly in line
```
jashapiro, line 160: Why not exactly inline? Would `position_identity()` work here?

scripts/FigS6_celldex-ref-comparison.R
```diff
@@ -0,0 +1,192 @@
...
+diagnosis_colors <- diagnosis_group_palette$color |> 
+  purrr::set_names(diagnosis_group_palette$diagnosis_group)
+
+# define suspension backgrounds to use on facet strips 
+backgrounds <- rep(c("Brain and CNS", "Leukemia", "Sarcoma"), each = 3) |>
+  purrr::map(\(x) element_rect(fill = diagnosis_colors[[x]]))
+
+# create delta median plot 
+all_delta_plot <- ggplot(all_delta_df, aes(x = celldex_reference, y = delta_median, shape = confident, alpha = confident)) +
+  ggforce::geom_sina(
```
jashapiro, line 156: The fact that the confident and non-confident distributions have different widths is a bit odd (makes sense, but looks funny). Maybe try `scale = "width"` and see how that looks?

I also wondered if we needed to worry about sorting to make the non-confident cells be on top. I'm not sure exactly whether it would make a difference, but adding a sort on `confident` to `all_delta_df` might be worth a look

jashapiro (approved):
I think this all looks good. I just had a couple small aesthetic questions about the plot.

Nothing that really needs another look from me.

scripts/FigS6_celldex-ref-comparison.R
allyhawkins: Just want to confirm that we like using `scale = "width"` as an option? I played around with sorting, but personally I think it all looked worse. 

Screenshot 2024-03-11 at 11 12 23 AM

jashapiro: I might make the width a bit narrower by setting `maxwidth=0.8` or something like that. 

But I realized that the main reason that it doesn't look as I expected is because `geom_sina` does not understand `shape` or `fill` as aesthetics, which means that you can't get them to use the same distribution unless the only difference is color. I like the open circles, so I am torn!

allyhawkins: I also like the open circles so I think we should keep that feature. Here's with `maxwidth=0.5`. I think I like this option the best, what do we think? 

Screenshot 2024-03-11 at 12 32 24 PM

PR #80, opened 2024-03-11
Fig S7 - CellAssign justification figure 

### LINKED ISSUE #41: Supplemental figure for how we chose cell type annotation methods

allyhawkins:
Closes #41 

This PR adds in the script to generate the supplemental figure summarizing the justification behind using `CellAssign`. Panel A is a UMAP summarizing the cell type annotations obtained from `CellAssign` and panel B shows the heatmap for comparing `CellAssign` to submitter-provided annotations. 

- I ended up using a different library than I noted in #41, because it turns out we were already using that library for Fig S5. I thought it would be strange to include the same library with almost the exact same heatmap in two places. I didn't combine with Fig. S5 since the focus there is inclusion of submitter provided annotations and comparing them with both cell type methods. 
- Instead of including 8 UMAPs like we do in the report, I only include 6 panels. I think that's sufficient, since most of the cells are accounted for in the top 2 cell types + Unknown.


scripts/FigS7_cellassign-justification.R
```diff
@@ -0,0 +1,152 @@
...
+  border = TRUE,
+  ## Row parameters
+  cluster_rows = FALSE,
+  row_title = "CellAssign annotations",
+  row_title_gp = grid::gpar(fontsize = 12),
+  row_title_side = "right",
+  row_names_side = "left",
+  row_names_gp = grid::gpar(fontsize = 10),
+  ## Column parameters
+  cluster_columns = FALSE,
```
jashapiro, line 129: I see that we aren't clustering, but I wonder if we want to look at the sorting of the input matrix for the submitter cell types only to group all of the tumor cell annotations?

jashapiro (approved):
This looks good to me, with only a couple of comments about sorting (One of which maybe applies to other heatmaps in the supplement?)

Approving, because it seems pretty minor, but if you want me to have another look, let me know.

scripts/FigS7_cellassign-justification.R
```diff
@@ -0,0 +1,152 @@
...
+# UMAP -------------------------------------------------------------------------
+# Create data frame of cell types
+celltype_df <- create_celltype_df(processed_sce)
+
+# lump celltypes for cellassign 
+celltype_df <- celltype_df |> 
+  dplyr::mutate(cellassign_celltype_annotation_lumped = cellassign_celltype_annotation |> 
+                  forcats::fct_lump_n(5, other_level = "All remaining cell types", ties.method = "first") |>
+                  forcats::fct_infreq() |>
+                  forcats::fct_relevel("Unknown cell type", "All remaining cell types", after = Inf)
```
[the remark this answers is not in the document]
allyhawkins: I was keeping this the same as we present it in the supplemental report, which means all remaining cell types are last. That probably doesn't matter a whole lot, but to me `All remaining cell types` includes all other cells that aren't shown in the other panels, including `Unknown`, so I would keep it last.

scripts/FigS7_cellassign-justification.R
allyhawkins: Screenshot 2024-03-12 at 11 36 38 AM

This is sorted so that everything with `tumor` is first and then all other cell types. @jashapiro is this what you had in mind?

PR #82, opened 2024-03-13
Compile new S4 and S5

### LINKED ISSUE #81: Compile Fig S6 and S7
We will need to compile the panels for Fig S6 (SingleR/celldex justification) and S7 (CellAssign justification). 

One thing to note is that the order of the supplemental figures for cell typing, might actually change slightly after https://github.com/AlexsLemonade/ScPCA-manuscript/issues/70 is addressed. We might want to wait to do this until the text has been altered.

allyhawkins:
Closes #81 

Here I'm adding in the compiled pdf/png versions of the celldex and cellassign figures. These will now appear earlier in the text than what was S4 and S5, so part of this PR is switching the names. 

- Fig S4 should have the celldex comparison
- Fig S5 should have the CellAssign example where A is the UMAPs and B is the heatmap comparing to submitter annotations 
- Fig S6 is the diagnostic plots 
- Fig S7 is the heatmap showing submitter annotations vs. SingleR and CellAssign 

One note is that I previously was saving the panel for the celldex comparisons to a png file, but it was going to be really hard to edit any fonts there, which was going to bother me if that was the only figure not in Arial. So, I updated this to export to a pdf and then edited the fonts before saving to `figure_s4.png` and `figure_s5.pdf`. I think the overall file size is still okay, even though its a lot of points. 

I also updated the `alpha` to try and make a little bit more contrast between the open and closed points, and I think this looks a tiny bit better.


jashapiro (approved):
LGTM!

PR #84, opened 2024-03-18
Update multiplexed overview figure

### LINKED ISSUE #83: Update multiplexed workflow
Right now we are showing HTO FASTQ as input to demultiplexing, but not as input to alevin-fry. This isn't entirely accurate, so we should adjust this to show that both the RNA and HTO FASTQ are used as input to alevin-fry. We might also want to show the HTO by cell counts matrix as we do with ADT.

allyhawkins:
Closes #83 

This PR makes an update to the multiplexed workflow to more accurately describe what's happening. The HTO FASTQ and RNA FASTQ should be input to both alevin-fry and demultiplexing. I also am showing the HTO by cell counts matrix as output from alevin-fry here.


sjspielman:
The main comment I have is if we want the "forked arrow" I put into this part of the ADT figure to also be in this one? I bet you could just drag down and replace the "read into R" arrow with that one?

allyhawkins:
> The main comment I have is if we want the "forked arrow" I put into this part of the ADT figure to also be in this one? I bet you could just drag down and replace the "read into R" arrow with that one?

Good call! I updated to include that arrow.

sjspielman (approved):
🌷

Issue #68, opened 2024-02-26
Release the figures


sjspielman:
Once we have completed figures, we'll want to tag a release which can be used to link the correct figure versions in `ScPCA-manuscript`.

Issue #33, opened 2024-01-18
Make README more public-facing before going public


sjspielman:
Before we make this repo public, we'll want to clean up some text in the README and `generate-figures-tables.sh` script comments to be public-facing. Presumably (?) this will also come hand-in-hand with public S3 bucket for figure data, or some other mechanism to distribute figure data if not.

Spots to update:

https://github.com/AlexsLemonade/scpca-paper-figures/blob/d927751e9ca62abb0af862a1dfe4ef0bf27c32af/README.md?plain=1#L48-L61

https://github.com/AlexsLemonade/scpca-paper-figures/blob/d927751e9ca62abb0af862a1dfe4ef0bf27c32af/generate-figures-tables.sh#L5

sjspielman:
Ugh I'm mad that I have a wrong verb in here (`which currently require S3` should be `requireS`, but not mad enough to change it since eventually this line will go away anyways!

allyhawkins:
To do this, we also need to move any data to a public S3 bucket.

sjspielman:
> To do this, we also need to move any data to a public S3 bucket.

We'll also need to update paths:

https://github.com/AlexsLemonade/scpca-paper-figures/blob/d927751e9ca62abb0af862a1dfe4ef0bf27c32af/scripts/figure_setup/sync-data-files.R#L7

https://github.com/AlexsLemonade/scpca-paper-figures/blob/d927751e9ca62abb0af862a1dfe4ef0bf27c32af/scripts/figure_setup/sync-metadata.R#L5-L6

jaclyn-taroni:
_If_ we were to make the data available outside the context of the Portal (i.e., without agreeing to Terms of Use), we would pare down the available fields in the sample metadata, most of which are not required for plotting.

My number one choice would be addressing https://github.com/AlexsLemonade/scpca-portal/issues/312 and then adapting the code for that.

sjspielman:
> Presumably (?) this will also come hand-in-hand with public S3 bucket for figure data, or some other mechanism to distribute figure data if not.

Any current thoughts here @jaclyn-taroni? We're not sending this out for _review_ at this time, but do we want to have something de-metadata'd data publicly available alongside a pre-print?

jaclyn-taroni:
I do not think that is necessary right now.

sjspielman:
Closing in favor of other issues which promote public-facing reproducibility.

PR #85, opened 2024-03-22
Update README.md to note internal access requirements

jaclyn-taroni:
Addresses #33 for now – enough to unblock #68 in my opinion.


allyhawkins (approved):
👍

PR #86, opened 2024-04-03
Add columns for 10X kits to modality table

allyhawkins:
Related to https://github.com/AlexsLemonade/ScPCA-manuscript/issues/108

This PR adds three new columns to the existing supplemental table summarizing the total number of samples and libraries across modalities. I added in a column for each of the 10X kits used, 10Xv2, 10Xv3, and 10Xv3.1. In the table, I put these immediately after the single-cell and single-nuclei columns and before any additional modality columns. 

I also deleted some extra code that wasn't being used anywhere in the script.

### DIFF
--- a/scripts/TableS1_modality-summary.R
+++ b/scripts/TableS1_modality-summary.R
@@ -43,11 +43,6 @@ sample_metadata_df <- readr::read_tsv(sample_metadata_file) |>
 
 # Create table -----------------------------------------------------------------
 
-# create a list of all the modality values
-# we will use this to total the number of libraries for a project across all modalities 
-# these will also correspond to the column names in the final table that contain the total for each modality 
-modality_columns <- c("Single-cell", "Single-nucleus", "Bulk RNA", "Spatial transcriptomics", "With CITE-seq", "With cell hashing")
-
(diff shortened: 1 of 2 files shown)


jaclyn-taroni (commented):
You have some NAs in `Diagnosis group` where you did not before: `SCPCP000013` and `SCPCP000012`

That appears to be because these are now split into two rows? I haven't done any diagnostics beyond that.

allyhawkins:
Ah it's because when we originally generated this we didn't have the new Dyer/Chen samples in the metadata. I need to update the script to remove any samples that are not on the portal before creating the table.

allyhawkins:
Okay I made a few updates to this to account for the errors in the table: 

- The individual project metadata files are the only place we have the `on_portal` column. So I added syncing those files in the `sync-metadata.R` script. 
- To create the table, I first create a sample whitelist by reading in each of the project metadata files and filtering to samples with `on_portal=TRUE`. 
- I also found an error where we were over counting the 10X kits for multiplexed libraries so I rearranged where that counting takes place to make sure we only count 1 per unique library id. 

This should now be ready for another look @jaclyn-taroni.

PR #87, opened 2024-04-16
Update sample numbers in figure panels and tables

allyhawkins:
Related to https://github.com/AlexsLemonade/ScPCA-manuscript/issues/90, this PR updates the panels in Figure 1 and modality table to include all projects and samples that are on the Portal. 

In #86, I introduced some code to make sure we are only displaying data for samples that are actually on the Portal. This excludes showing anything for samples that we currently have but are not yet on the Portal. To update the figures here I needed to also create a sample whitelist of samples that are on the portal, so I took out that code snippet and turned it into a helper function that is used across the scripts for sample and modality overviews. 

I added SCPCP000020 and SCPCP000021 to the project whitelist, so they should now be included in the modality table and bar chart. For the sample summary overview (Fig 1A), we only were including patient or PDX samples and not cell lines. I added a line that removes any samples that are cell lines so they are not included in that plot, but they are included in anything library related. 

Just a note that we lost one Leukemia sample because we removed one per request of the submitter since generating this. We also lost a Wilms tumor in the overall count, which I can't quite figure out. However, I did check that the number is as expected based on what's in the Portal and 53 is correct. Note that the leukemia sample was a single-cell sample so that's why that number is lower in the modality bar chart. 

Once this is approved then I will update the actual figure. I want to make sure this is good before spending time in illustrator making sure the figure looks good.


scripts/utils/sample-summary-helper-functions.R
```diff
@@ -0,0 +1,28 @@
+
+#' Get whitelist of samples to include in plots 
+#'
+#' @param project_metadata_file Path to file containing ScPCA project metadata
```
sjspielman, line 4: need to document the `project_whitelist` argument

scripts/utils/sample-summary-helper-functions.R
```diff
@@ -0,0 +1,28 @@
...
+      sample_list <- readr::read_tsv(file) |> 
+        dplyr::filter(on_portal) |> 
+        dplyr::pull(scpca_sample_id)
+      
+      return(sample_list)
+    }) |> 
+    unlist() 
+  
+  return(sample_whitelist)
+}
```
sjspielman, line 28:
```suggestion
}

```

figures/pngs/Fig1B_modality-summary.png
sjspielman, line 1: Do we still need this PNG file? Should we remove the code that exports it and the file itself?

figures/pdfs/Fig1_sample-summary.pdf
sjspielman, line 1: The y-axis text for the brain panel is a little bit tight now that we have more diagnoses along the axis. Can you make the `axis.text.y` a bit smaller?

sjspielman (commented):
This mostly all looks good to me! Left a couple small comments to address.

manuscript-numbers/diagnosis-group-counts.tsv
sjspielman, line 1: Hm actually while reviewing #119 I realized two of these numbers went _down_ by 1...? Did we expect that?

allyhawkins: > Just a note that we lost one Leukemia sample because we removed one per request of the submitter since generating this. We also lost a Wilms tumor in the overall count, which I can't quite figure out. However, I did check that the number is as expected based on what's in the Portal and 53 is correct. Note that the leukemia sample was a single-cell sample so that's why that number is lower in the modality bar chart.

So yes we expected it for leukemia, for the "other" group its a wilms tumor and I don't know why but I did double check that this number is correct.

allyhawkins:
@sjspielman I addressed your comments, including changing the font size. Although, it's all going to change 
when I actually compile the figure so I don't want to spend too much time on that.

sjspielman (approved):
👍

Issue #99, opened 2025-01-06
Prepare TPM tables for bulk data


sjspielman:
_Broken out of #93_

As the first step to performing deconvolution, we need to prepare TPM tables for input to method(s). The input information needed for this is in the `quant.sf` files from `salmon` in our S3 bucket checkpoint files. This issue will also necessarily establish a directory in this repository where this deconvolution analysis will be stored. Associated data and results will also be "stored" in this directory, but licensed ScPCA data will be `.gitignore`'d.

This will require several steps/scripts:

0. [Analysis setup script] Script to obtain `quant.sf` files locally. This script will be, in spirit/concept, similar to our [data syncing script](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/scripts/figure_setup/sync-data-files.R). When downloaded, these files will be saved (but again, `.gitignore`'d!) in this new directory according to a `project/sample/library` structure. Since these files are not organized in this manner on S3 (as they are checkpoint), we'll need a library metadata file linking ids to support this.
1. Script to calculate TPM tables. This script should take a project ID as input, and export a single TSV file with TPM values for that project id.  We should use `tximport` to calculate TPM.
2. Shell script to serve as a deconvolution "workflow." It should loop over projects and the first step will be TPM calculations.

This may end up being 2 PRs: 
1. Establish analysis directory and write setup script to obtain `quant.sf`
2. Write scripts for calculating TPM

allyhawkins:
> Since these files are not organized in this manner on S3 (as they are checkpoint), we'll need a library metadata file linking ids to support this.

Just a note that we have a [script that syncs the metadata files already](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/scripts/figure_setup/sync-metadata.R), so you should be able to use the library and sample metadata as needed for syncing the `quant.sf` files.

sjspielman:
One additional aspect: The deconvolution methods we've seen so far require gene symbols, not ensembl IDs. In my own code to prep here (#96 & #97) I did some very quick-and-dirty conversion here using SCE colData to map ids. However, I realize this approach does not handle duplicates very well, and there are indeed duplicates symbols which come out of quick conversions. For the paper itself, we'd like to perform this conversion a little more intelligently!

[...]

PR #102, opened 2025-01-06
Establish deconvolution directory and sync salmon files

sjspielman:
Part 1 of #99 

This PR establishes a directory `analysis/bulk-deconvolution` for the, you guessed it, bulk deconvolution analysis. To keep open the (probably slim but non-zero) possibility that we might do complementary/other scientific analysis here, I decided to nest this in an `analysis` directory where future science can live. 

This directory contains readmes,  `scripts`, and `data` (mostly ignored) directory, and currently one script to sync the quant.sf files. This script uses existing (if present, stops otherwise) metadata files to determine the files to sync, and then syncs to `data/salmon-quant-files` organized in the "usual" project/sample/library order. 

Finally, I also updated the `renv::lock` file by running `renv::restore()` and pretty much agreeing to what it wanted me to agree to in order to recreate the environment! Mostly R was bumped to 4.3.3, and package versions/hashes along with it. It also CRAN-ified some of the RSPMs, which can be changed back as needed.

The next PR after this will actually do TPM calculations and export TSVs into `data/tpm/`.

### DIFF
--- a/renv.lock
+++ b/renv.lock
@@ -1,6 +1,6 @@
 {
   "R": {
-    "Version": "4.3.1",
+    "Version": "4.3.3",
     "Repositories": [
       {
         "Name": "BioCsoft",
@@ -132,6 +132,7 @@
       "Package": "BiocVersion",
       "Version": "3.18.1",
       "Source": "Bioconductor",
+      "Repository": "Bioconductor 3.18",
       "Requirements": [
         "R"
       ],
@@ -178,14 +179,14 @@
     },
     "DBI": {
       "Package": "DBI",
-      "Version": "1.2.2",
+      "Version": "1.2.1",
       "Source": "Repository",
       "Repository": "RSPM",
       "Requirements": [
         "R",
         "methods"
       ],
-      "Hash": "164809cd72e1d5160b4cb3aa57f510fe"
+      "Hash": "9b4993e98e0e19da84c168460c032fef"
     },
     "DelayedArray": {
       "Package": "DelayedArray",
@@ -264,7 +265,7 @@
     },
     "GenomeInfoDb": {
       "Package": "GenomeInfoDb",
-      "Version": "1.38.6",
+      "Version": "1.38.5",
       "Source": "Bioconductor",
       "Repository": "Bioconductor 3.18",
       "Requirements": [
@@ -279,7 +280,7 @@
         "stats4",
         "utils"
       ],
-      "Hash": "7d566eec72f05303e8e799cebc49d334"
+      "Hash": "7575d02551088d69cfcbf65a3a6f0b08"
     },
     "GenomeInfoDbData": {
       "Package": "GenomeInfoDbData",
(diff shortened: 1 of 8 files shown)


renv.lock
```diff
@@ -178,14 +179,14 @@
     },
     "DBI": {
       "Package": "DBI",
-      "Version": "1.2.2",
+      "Version": "1.2.1",
```
jashapiro, line 182: I would not expect downgrades like this with a restore. There are many of these, so I suspect something went a bit wrong with your restore.

sjspielman: Pretty sure I fixed it and now it's at renv `1.0.11` too. Worth noting that `tximport` now has a GitHub remove and no longer Bioconductor, presumably because this is a dependency change for some other package? Not sure why else that would have been swapped out.

jashapiro: I'm still unsure why you have so many changes to packages here with a restore. I ran a restore locally based on the `main` branch and ran into only a couple of issues: One was installing `HDF5Array` where I ran into a compilation issue so I did a manual install to update to a version with a binary package available, and the other was a complaint when I ran `renv::snapshot()` where an update to `Matrix` seemed to be required for compatibility.

The resulting `renv.lock` file is attached: 
[renv.lock.zip](https://github.com/user-attachments/files/18323214/renv.lock.zip)

I'm not opposed to upgrading more, but I don't want to do it without some checking of the other code that might be involved, and if we are going to do updates, it probably makes sense to update to the version of `scpcaTools` that is actually used in the workflow (0.4.1), which would actually be a more significant update, as we would also be updating to R 4.4 and Bioc 3.19.

Which is to say that I would probably remove all renv changes from this PR and file a separate PR with a more complete renv update.

sjspielman: > I'm still unsure why you have so many changes to packages here with a restore.

jashapiro (approved):
LGTM

You will probably want to send the renv update PR to @allyhawkins, as she is best positioned to check that the other scripts still run as expected.

sjspielman:
> You will probably want to send the renv update PR to @allyhawkins, as she is best positioned to check that the other scripts still run as expected.

@allyhawkins fyi I have some around to very much thinking the problem was my cache and not the lock file...I just had forgotten the magic of `rebuild=TRUE` 🎉. So you may have one less review to do there :)

jashapiro:
> > You will probably want to send the renv update PR to @allyhawkins, as she is best positioned to check that the other scripts still run as expected.
> 
> @allyhawkins fyi I have come* around to very much thinking the problem was my cache and not the lock file...I just had forgotten the magic of `rebuild=TRUE` 🎉. So you may have one less review to do there :)

I still don't get it, as `restore()` should still be matching the lockfile, not the versions you have already installed. That is what it is for! `rebuild` should only affect packages where you already have a cached copy of the same version. 

But separately, I do still think that matching the R/Bioc and `scpcaTools` versions across repos is probably worth doing!

PR #105, opened 2025-01-06
Add script to calculate TPM

### LINKED ISSUE #99: Prepare TPM tables for bulk data

sjspielman:
Closes #99 

This PR adds two scripts:
- A script to calculate TPM for a given project and export a TSV file of results
  - I use an SCE to help id mapping, and I sum the duplicates as discussed. I use tidyverse and export a tsv here, but there's a good argument to use base R and matrices instead, especially since we'll have to convert these to matrices anyways! But, I enjoy tidyverse so that's what I did. Let me know if we prefer I change that.
  - A shell script for the analysis itself
    - This currently has a step with a _minimal check_ to download the `quant.sf` files - should this check be more involved? 
    - And then, we loop over project directories to calculate a TPM matrix for each. Next up is actual deconvolution.

There are several other updates:
- I updated readmes
- I updated `scripts/figure_setup/sync-reference-files.R` to also download the t2g file, since this script was doing those sorts of things anyways! I can move this sync into the deconvolution directory if preferred, though.
- I did have to get `optparse` into `renv`, so along the way that was indeed bumped to 1.0.5 which seemed not bad? I also saw this message go by: 
```
The following package(s) have unsatisfied dependencies:
- MatrixModels requires Matrix (>= 1.6-0), but version 1.5-4.1 is installed
Consider updating the required dependencies as appropriate.
```
So, I went ahead and bumped just that package.

### DIFF
--- a/analysis/bulk-deconvolution/run-deconvolution.sh
+++ b/analysis/bulk-deconvolution/run-deconvolution.sh
@@ -31,4 +31,3 @@ for project_dir in $salmon_quant_dir/*; do
     # forthcoming!
 
 done
-
--- a/analysis/bulk-deconvolution/run-deconvolution.sh
+++ b/analysis/bulk-deconvolution/run-deconvolution.sh
@@ -1,7 +1,6 @@
 #!/bin/bash
 
-# This script runs this analysis and I will add more comments later.
-# If you are using 1P, you may need op run -- .
+# This script runs the bulk deconvolution analysis.
 
 set -euo pipefail
 
--- a/analysis/bulk-deconvolution/run-deconvolution.sh
+++ b/analysis/bulk-deconvolution/run-deconvolution.sh
@@ -12,11 +12,12 @@ cd "$basedir"
 data_dir="data"
 script_dir="scripts"
 salmon_quant_dir="${data_dir}/salmon-quant-files"
+reference_dir=${data_dir}/reference # stores the id map file
 tpm_dir="${data_dir}/tpm"
 
-# Step 0: Download quant.sf files if they do not exist
-if [ ! -d $salmon_quant_dir ]; then
-    Rscript ${script_dir}/figure_setup/sync-salmon-output.R
+# Step 0: Prepare id map and download quant.sf files if they do not exist
+if [[ ! -d $salmon_quant_dir || ! -d $reference_dir ]]; then
+    Rscript ${script_dir}/figure_setup/prepare-data-files.R
 fi
 
 for project_dir in $salmon_quant_dir/*; do
--- a/analysis/bulk-deconvolution/run-deconvolution.sh
+++ b/analysis/bulk-deconvolution/run-deconvolution.sh
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# This script runs this analysis and I will add more comments later.
+# If you are using 1P, you may need op run -- .
+
+set -euo pipefail
+
+# Run script from its location
+basedir=$(dirname "${BASH_SOURCE[0]}")
+cd "$basedir"
+
+# Define directories
+data_dir="data"
+script_dir="scripts"
+salmon_quant_dir="${data_dir}/salmon-quant-files"
+tpm_dir="${data_dir}/tpm"
+
+# Step 0: Download quant.sf files if they do not exist
+if [ ! -d $salmon_quant_dir ]; then
+    Rscript ${script_dir}/figure_setup/sync-salmon-output.R
+fi
+
+for project_dir in $salmon_quant_dir/*; do
+    project_id=$(basename $project_dir)
+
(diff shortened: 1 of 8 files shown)


analysis/bulk-deconvolution/scripts/calculate-tpm.R
```diff
@@ -0,0 +1,124 @@
...
+      # return data frame where the tpm column is named by the sample id
+      sample_tpm_df <- data.frame(tpm = tpm)
+      names(sample_tpm_df) <- sample_id
+      sample_tpm_df
+
+    }
+  ) |>
+  # we expect the same genes all around here since quantification used the same gene set
+  dplyr::bind_cols() |>
+  tibble::rownames_to_column(var = "ensembl_id")
```
jashapiro, line 107: You can let `tximport` do all of the table making here: you just need `quant_files` to be named with the sample_ids
```suggestion

quant_names <- stringr::str_split_i(quant_files, pattern = "/", i = 2)
quant_paths <- setNames(file.path(data_dir, quant_files), quant_names)


txi_salmon <- tximport::tximport(
  quant_paths,
  type = "salmon",
  tx2gene = t2g_table
)

tpm_mat  <- txi_salmon$abundance
```

Then `rowsum()` can replace all of the tidyverse stuff:

```
symbol_map <- setNames(map_table$gene_symbol, map_table$ensembl_id)
tpm_summed_mat <- rowsum(
  tpm_mat,
  group = symbol_map[rownames(tpm_mat)]
)
```


The only other thing I might suggest is that you _could_ avoid loading the tx2gene info by using `tximeta` but since we already have all that done, there isn't too much point. Though it should get you the symbol translations too, without having to load a whole SCE object.

sjspielman: >  But you might move getting the translation table from an SCE to the previous script, just to prevent loading the whole SCE for each project.

And export a TSV of the "map_table" to read in here, you mean?

analysis/bulk-deconvolution/scripts/calculate-tpm.R
```diff
@@ -0,0 +1,100 @@
...
+# remove rows with no symbol
+tpm_mat_pruned <- tpm_mat[!is.na(symbol_map[rownames(tpm_mat)]), ]
+# sum across duplicated gene symbols
+tpm_summed_mat <- rowsum(
+  tpm_mat_pruned,
+  group = symbol_map[rownames(tpm_mat_pruned)]
+) # --> 39133 rows
+
+# Export to rds -------
+readr::write_rds(tpm_summed_mat, output_file)
```
jashapiro, line 100: I'm of two minds about writing as a tsv vs. rds here, but this probably makes sense.

analysis/bulk-deconvolution/run-deconvolution.sh
```diff
@@ -0,0 +1,34 @@
...
+data_dir="data"
+script_dir="scripts"
+salmon_quant_dir="${data_dir}/salmon-quant-files"
+reference_dir=${data_dir}/reference # stores the id map file
+tpm_dir="${data_dir}/tpm"
+
+# Step 0: Prepare id map and download quant.sf files if they do not exist
+if [[ ! -d $salmon_quant_dir || ! -d $reference_dir ]]; then
+    Rscript ${script_dir}/figure_setup/prepare-data-files.R
+fi
```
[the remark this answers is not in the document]
sjspielman: >  I might just make this a manual step separate from this script

Issue #90, opened 2024-12-13
Regenerate Figure 1 and supplementary tables with updated sample numbers 


allyhawkins:
We will need to update the following figures/tables to reflect the most recent sample numbers: 

- [x] 1A - diagnosis summary 
- [x] 1B - modality summary 
- [x] Table S1 - overview of portal datasets 
- [x] Table S2 - reference summary, add new projects 

To do this we should first add the new projects to the [`project-whitelist.txt`](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/sample-info/project-whitelist.txt) file. Then we should be able to run the `generate-figures-tables.sh` script to regenerate all of the above panels/tables. 

We will also need to update the compiled version of Figure 1 with the new panels.

sjspielman:
@allyhawkins It seems to me that some (all) of these items were accomplished in #106? How close is this to closing out? Asking for the purposes of re-making fig 1.

allyhawkins:
> [@allyhawkins](https://github.com/allyhawkins) It seems to me that some (all) of these items were accomplished in [#106](https://github.com/AlexsLemonade/scpca-paper-figures/pull/106)? How close is this to closing out? Asking for the purposes of re-making fig 1.

These are all done. I actually had on my list today to recompile the figure.

sjspielman:
> I actually had on my list today to recompile the figure.

Aha, cool cool. So, I'll close out #114 and we'll let this issue handle that recompilation.

PR #106, opened 2025-01-07
Update sample numbers for Figure 1 and supplemental tables 

allyhawkins:
Addresses #90 part 1 

Here I am updating the panels for Fig 1A and 1B and the supplemental tables that include an overview of the sample types on the Portal. I am not yet including the actual compilation of the new Figure 1 here because there are some more changes than I anticipated and I want to make sure we are on the same page with these updates before making the changes to the compiled figure. 

- I added all the new projects that we have added to our project whitelist. 
- I updated the diagnosis groupings and disease timing reference tables we have to account for new samples. 
  - There are 3 new diagnoses that we did not have before: hepatoblastoma, melanoma, and high-grade sarcoma. Hepatoblastoma and melanoma were pretty easy to add since we had the `MONDO` id for these already and they both belong in the "other" group. However, we don't yet have a `MONDO` id for the high-grade sarcoma samples, which is something we should add to the sample metadata. Once we settle on a term here I'll file tickets to make those updates accordingly. On that note, there is no clear term for "high-grade sarcoma" since it really encompasses a variety of different sarcomas, so I just used the term for [sarcoma](https://www.ebi.ac.uk/ols4/ontologies/mondo/classes/http%253A%252F%252Fpurl.obolibrary.org%252Fobo%252FMONDO_0005089?lang=en). This really doesn't affect the figures, but we include it in the table so I thought I would add it. 
  - There is a new disease timing value, `Remission`, that is being added. This only affects one sample, so I didn't think it needed its own group. I included it in the `During or after treatment` group. On that same note, that group was never included in the bar plot before since none of those samples were actually on the Portal so it got filtered out. This means we have one more group that we need a pattern for. I could not find any patterns that worked with the set up we had and could be distinguishable (I remember this being very annoying when first creating this figure). So to keep the same patterns we have I decided to group "Progressive" and "Recurrence" samples into the same pattern group. How do we feel about this? 
- I also updated the two supplemental tables here and made a small code change to ensure that only projects that were on the portal got included in the cell type table since that was missing.

### DIFF
--- a/renv.lock
+++ b/renv.lock
@@ -931,6 +931,39 @@
       ],
       "Hash": "8fcc818f3b9887aebaf206f141437cc9"
     },
+    "bslib": {
+      "Package": "bslib",
+      "Version": "0.8.0",
+      "Source": "Repository",
+      "Repository": "RSPM",
+      "Requirements": [
+        "R",
+        "base64enc",
+        "cachem",
+        "fastmap",
+        "grDevices",
+        "htmltools",
+        "jquerylib",
+        "jsonlite",
+        "lifecycle",
+        "memoise",
+        "mime",
+        "rlang",
+        "sass"
+      ],
+      "Hash": "b299c6741ca9746fb227debcb0f9fb6c"
+    },
+    "cachem": {
+      "Package": "cachem",
+      "Version": "1.1.0",
+      "Source": "Repository",
+      "Repository": "RSPM",
+      "Requirements": [
+        "fastmap",
+        "rlang"
+      ],
+      "Hash": "cd9a672193789068eb5a2aad65a0dedf"
+    },
     "car": {
       "Package": "car",
       "Version": "3.1-3",
@@ -1244,6 +1277,13 @@
       "Source": "Repository",
       "Repository": "RSPM",
       "Hash": "680887028577f3fa2a81e410ed0d6e42"
+    },
+    "fastmap": {
+      "Package": "fastmap",
+      "Version": "1.2.0",
+      "Source": "Repository",
+      "Repository": "RSPM",
+      "Hash": "aa5e1cd11c2d15497494c5292d7ffcc8"
     },
     "fishpond": {
       "Package": "fishpond",
@@ -1923,6 +1963,17 @@
       ],
       "Hash": "8885ffb1f46e820dede6b2ca9442abca"
     },
+    "memoise": {
+      "Package": "memoise",
+      "Version": "2.0.1",
+      "Source": "Repository",
+      "Repository": "RSPM",
+      "Requirements": [
+        "cachem",
+        "rlang"
+      ],
+      "Hash": "e2817ccf4a065c5d9d7f2cfbe7c1d78c"
+    },
     "mgcv": {
       "Package": "mgcv",
       "Version": "1.9-1",
@@ -2248,18 +2299,6 @@
       ],
       "Hash": "f4625e061cb2865f111b47ff163a5ca6"
     },
-    "proxy": {
-      "Package": "proxy",
-      "Version": "0.4-27",
-      "Source": "Repository",
-      "Repository": "CRAN",
-      "Requirements": [
-        "R",
-        "stats",
-        "utils"
-      ],
-      "Hash": "e0ef355c12942cf7a6b91a6cfaea8b3e"
-    },
     "purrr": {
       "Package": "purrr",
       "Version": "1.0.2",
@@ -2515,6 +2554,20 @@
         "R"
       ],
       "Hash": "b462187d887abc519894874486dbd6fd"
+    },
+    "sass": {
+      "Package": "sass",
+      "Version": "0.4.9",
+      "Source": "Repository",
+      "Repository": "RSPM",
+      "Requirements": [
+        "R6",
+        "fs",
+        "htmltools",
+        "rappdirs",
+        "rlang"
+      ],
+      "Hash": "d53dbfddf695303ea4ad66f86e99b95d"
     },
     "scales": {
       "Package": "scales",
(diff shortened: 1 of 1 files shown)


allyhawkins:
@jaclyn-taroni I'm working on updating this since the new renv updates have gone in (see #108). We now have a new version of `ggpattern` and the `wave` pattern we were previously using doesn't seem to be supported... or at least the code is not working in the same way it was before and that pattern is blank in the figure. I've been trying to find a solution with no luck so far, so no need to review this just yet. I'll tag you when it's ready.

allyhawkins:
@jaclyn-taroni I've done a lot of playing around with figure 1A since `ggpattern` no longer seems to support the `wave` option with `ggpattern::geom_bar_pattern`. I've been struggling with `ggpattern` because there really are only 3 pattern options now, `circle`, `crosshatch`, and `stripe` (along with `none`) that work for [bar charts](https://coolbutuseless.github.io/package/ggpattern/articles/geom-gallery-geometry.html). 

That means we would have to condense our groups to 4 disease timing options rather than the 5 that we had prior to the new updates. Now with the new samples, we have a new group (during or after treatment) that make this number go up to 6. I tried playing around with adjusting the pattern and the angle, but it was just all pretty bad and when you do both then we have two different legends and it was hard to see differences.  

So I'm thinking maybe we move away from using `ggpattern` here and use colors for the disease timing rather than the diagnosis group. I made an example of what I think this would look like using colors from the `Okabe-Ito` palette. I still need to adjust the ordering in the legend, but what do you think of this rather than the patterns? 

image

For reference this is the old plot: 
https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/figures/pdfs/Fig1_sample-summary.pdf

If you like this then I'll make the necessary code changes here.

jashapiro:
This seems like a bit of a bug rather than an intentional change, but you can work around it by defining the fills directly with `gridpattern::patternFill()` and then applying them using a `ggplot2::geom_bar()` with `scale_fill_manual`. (ggplot2 added pattern support in 3.5, it seems: https://trevorldavis.com/R/gridpattern/#using-these-patterns-with-the-ggplot2-package)

Unfortunately, it seems like you can't control the colors and patterns separately, which makes it harder to reproduce the old plot. It is still probably doable with overlaying a patterned but transparent plot over the colored bars, but I don't know if you want to go that far...

jashapiro:
Update: I think if you add `pattern_frequency` to the call, `wave` should work again.  `https://stackoverflow.com/questions/79017327/wave-pattern-in-ggpattern-is-not-working`

allyhawkins:
> Update: I think if you add `pattern_frequency` to the call, `wave` should work again: https://stackoverflow.com/questions/79017327/wave-pattern-in-ggpattern-is-not-working

That does appear to work, thanks @jashapiro! 

image

With the pattern option we still need to combine two classes since we only have 5 patterns so if we want to separate out "Progressive" and "Recurrence" then we should use the colors instead. I personally like keeping them separate, but I'll let @jaclyn-taroni make the call.

jashapiro:
> > Update: I think if you add `pattern_frequency` to the call, `wave` should work again: https://stackoverflow.com/questions/79017327/wave-pattern-in-ggpattern-is-not-working
> 
> That does appear to work, thanks @jashapiro!
> 

There is still something odd there with the patterns: they are quite different between the actual bars and the legend, and I have no idea how to fix that (I'm looking at the osteosarcoma bar).

jaclyn-taroni:
I would prefer not to combine Progressive and Recurrence into a single category, so it sounds like that means we need to use colors instead of patterns.

allyhawkins:
@jaclyn-taroni I've updated this to use colors for Fig1A and remove the use of `ggpattern`. This is now ready for review.

palettes/disease-timing-palette.tsv
```diff
@@ -0,0 +1,7 @@
+disease_timing	color
+Initial diagnosis	#FFFFFF
+During or after treatment	#CC79A7
+Progressive	#E69F00
+Recurrence	#0072B2
+Post-mortem	#009E73
+Unknown	#000000
```
jaclyn-taroni, line 7: New line

```suggestion
Unknown	#000000

```

renv.lock
jaclyn-taroni, line 1: Can you catch me up on why I shouldn't be worried about all these removals?

sample-info/diagnosis-groupings.tsv
```diff
@@ -45,6 +47,7 @@ MONDO:0019373	Desmoplastic small round cell tumor	Desmoplastic small round cell
 MONDO:0015795	Undifferentiated embryonal sarcoma of the liver	Embryonal sarcoma	Sarcoma
 MONDO:0017387	Epithelioid sarcoma	Epithelioid sarcoma	Sarcoma
 MONDO:0012817	Ewing sarcoma	Ewing sarcoma	Sarcoma
+MONDO:0005089	sarcoma	High-grade sarcoma	Sarcoma
```
jaclyn-taroni, line 50: This seems like a reasonable choice of term.

jaclyn-taroni (approved):
I'm not totally up to date with what's going on with `renv` with this repository, so if all the removals in the `renv.lock` file are not a concern, feel free to merge.

**Edit:** Now I see there are conflicts with `renv.lock`.

renv.lock
allyhawkins: `ggpattern` and it's dependencies got removed when I updated the code here. I just merged in the changes from `main`, keeping any packages that were added in `main` (which is where the conflicts were) and then ran `renv::restore()` and `renv::snapshot()` locally to make sure everything was all set. So I believe these removals are all from `ggpattern` and code changes here.

Issue #107, opened 2025-01-07
Regenerate figures with updated dependencies 


sjspielman:
Following #90 & #104, we'll want to regenerate _all_ figures with the updated `renv.lock`. There are a few changes we may expect beyond what is noted in #90 due to updates to `scpca-nf` itself since we originally generated the figures, so we should confirm any changes are indeed expected.

PR #108, opened 2025-01-07
Update renv.lock and fix scripts accordingly

### LINKED ISSUE #104: Update dependency versions
From https://github.com/AlexsLemonade/scpca-paper-figures/pull/102#discussion_r1904511020, 

> it probably makes sense to update to the version of scpcaTools that is actually used in the workflow (0.4.1), which would actually be a more significant update, as we would also be updating to R 4.4 and Bioc 3.19.

This issue tracks a general version bump for overall consistency among ScPCA happenings.

sjspielman:
Closes #104 

This PR updates the `renv.lock` to use BioC 3.19, R 4.4.0, and scpcaTools 0.4.1. Other package versions were updated accordingly. I did this by running `renv::init("bioconductor@3.19"); renv::snapshot()` from an RStudio session running R 4.4.0, and things just kind of worked(!). I did then override the renv 1.0.11 version which got installed to instead use 1.0.10 since 1.0.11 has a known bug that has proven annoying (it probably wouldn't be a problem in this repo, but it was really annoying when we encountered it, so I have been burned.... that [bug](https://github.com/rstudio/renv/issues/2017) has been fixed in dev, but it hasn't made it to CRAN yet.).

I also ran the figure/table code and fixed a few spots which needed to be made compatible with the updated version of ggplot2. I did _not_ commit any of updated figures themselves, just their scripts! I opened a separate issue (#107) to formally regenerate all figures after this and other updates go in. This PR does ensure that all scripts run to prepare for official regeneration, though!

For the code changes, you'll notice that I nudged some legends down to a height of 0.88. This is because the updated genes detected vs. UMI count plot had a point interfering with the legend location where it previously was at 0.9, so I lowered it and also made other panels consistent at 0.88.


sjspielman:
I'm requesting @allyhawkins here for review since you have the most familiarity with the figure code here, and also you're working on figures in parallel so you should have some say over how this PR gets handled amid your code changes :)

jashapiro:
I would suggest setting the R and bioconductor versions in `settings.json` as well; as that should prevent them from being overridden in the lockfile (notably by people running R 4.4.x, which should not cause trouble, but is _different_).

renv.lock
```diff
@@ -2523,19 +2534,18 @@
...
       "Source": "GitHub",
       "RemoteType": "github",
-      "Remotes": "AlexsLemonade/scpcaData, immunogenomics/lisi@v1.0",
       "RemoteHost": "api.github.com",
-      "RemoteRepo": "scpcaTools",
       "RemoteUsername": "AlexsLemonade",
-      "RemoteRef": "HEAD",
-      "RemoteSha": "2ebdc4f4dfc4233fad97805f9a9a5e3bc6919f1e",
+      "RemoteRepo": "scpcaTools",
+      "RemoteRef": "main",
```
jashapiro, line 2547: I think we probably want to peg this?

This could also be done with `renv::install("AlexsLemonade/scpcaTools@v0.4.1")`

```suggestion
      "RemoteRef": "v0.4.1",
```

sjspielman:
> I would suggest setting the R and bioconductor versions in renv/settings.json

I did set the bioconductor version, but not R. Will do.

renv.lock
sjspielman: ✅ https://github.com/AlexsLemonade/scpca-paper-figures/pull/108/commits/37c39c9102d109c7feedcbf6ee074960fd079713

sjspielman:
@jashapiro do you have any thoughts about the `ppm` settings in the json? I have ppm in the lockfile itself, but perhaps telling settings about it will prevent cran from engaging in a hostile takeover which seems to happen sometimes(?)

jashapiro:
> @jashapiro do you have any thoughts about the `ppm` settings in the json? I have ppm in the lockfile itself, but perhaps telling settings about it will prevent cran from engaging in a hostile takeover which seems to happen sometimes(?)

I don't think it is worth worrying about. This has more to do with how the packages are labeled when downloaded and cached and maybe something to do with default settings in `options("repos")` and/or RStudio packages pane.

allyhawkins:
Just a note that running `renv::restore()` throws an error for me: 

```
Error: Error installing package 'Cairo':
=================================

* installing *source* package ‘Cairo’ ...
** package ‘Cairo’ successfully unpacked and MD5 sums checked
** using staged installation
checking for gcc... clang -arch arm64
checking whether the C compiler works... yes
checking for C compiler default output file name... a.out
checking for suffix of executables... 
checking whether we are cross compiling... no
checking for suffix of object files... o
checking whether the compiler supports GNU C... yes
checking whether clang -arch arm64 accepts -g... yes
checking for clang -arch arm64 option to enable C11 features... none needed
checking for sys/wait.h that is POSIX.1 compatible... yes
checking for stdio.h... yes
checking for stdlib.h... yes
checking for string.h... yes
checking for inttypes.h... yes
checking for stdint.h... yes
checking for strings.h... yes
checking for sys/stat.h... yes
checking for sys/types.h... yes
checking for unistd.h... yes
checking for string.h... (cached) yes
checking for sys/time.h... yes
checking for unistd.h... (cached) yes
checking for an ANSI C-conforming const... yes
checking for pkg-config... /usr/local/bin/pkg-config
checking whether pkg-config knows about cairo... yes
checking for configurable backends... cairo cairo-ft cairo-pdf cairo-png cairo-ps cairo-xlib cairo-xlib-xrender
checking whether --static is needed... neither works, abandoning pkg-config
configure: CAIRO_LIBS is unset, attempting to guess it.
configure: CAIRO_CFLAGS=-I/usr/local/include/cairo
checking if R was compiled with the RConn patch... no
checking for cairo.h... yes
checking for PNG support in Cairo... yes
checking for ATS font support in Cairo... no
configure: CAIRO_LIBS=-L/usr/local/lib -lcairo
checking for library containing deflate... -lz
checking whether Cairo programs can be compiled... configure: error: Cannot compile a simple Cairo program. See config.log for details.
ERROR: configuration failed for package ‘Cairo’
* removing ‘/Users/allyhawkins/Documents/ALSF/git_repos/scpca-paper-figures/renv/staging/1/Cairo’
install of package 'Cairo' failed [error code 1]
```

I had tried updating this yesterday when I was having other `renv` issues and had this issue and I have not yet found a solution... I installed `cairo` on my computer and I've also tried to delete the `renv` folder and start from scratch and still having this issue.

jashapiro:
> Just a note that running `renv::restore()` throws an error for me:
> 
> ```
> Error: Error installing package 'Cairo':
> =================================
> 
> * installing *source* package ‘Cairo’ ...


When I run into problems like this, it is often because the P3M repository does not have a binary version, which forces a source install, which fails because of some difficult dependency. (I am assuming this is on your Mac, not the server?) The solution I have found often works is to install manually, forcing direct loading from CRAN:

```
renv::install("Cairo", repos = "https://cloud.r-project.org")
```

Often this will work to force the binary install, and then things go smoothly afterward. (This works as long as the lock version is the most recent, as it is here.)

jashapiro:
> I was able to run this first and then run `renv::restore()` successfully! Should we update the lock file to install `Cairo` from `CRAN` rather than from `RSPM`?

If you run `renv::snapshot()` does it want to change anything? In general, the repo markers are informational and are not used for installation directly. We might be able to add `cloud.r-rpoject.org` as a fallback repo, but I'm not sure it would help in this case, unfortunately.

allyhawkins:
> > I was able to run this first and then run `renv::restore()` successfully! Should we update the lock file to install `Cairo` from `CRAN` rather than from `RSPM`?
> 
> If you run `renv::snapshot()` does it want to change anything? In general, the repo markers are informational and are not used for installation directly. We might be able to add `cloud.r-rpoject.org` as a fallback repo, but I'm not sure it would help in this case, unfortunately.

Running `renv::snapshot()` tells me the lock file is up to date, so I guess we're good!

allyhawkins (approved):
LGTM!

sjspielman:
@allyhawkins to confirm, are you good with your other code if I merge this now? Given Jackie's schedule, I think this is probably very safe since she won't likely get to review #106 for a bit more time yet, so there is time to sync that branch up.

allyhawkins:
> @allyhawkins to confirm, are you good with your other code if I merge this now? Given Jackie's schedule, I think this is probably very safe since she won't likely get to review #106 for a bit more time yet, so there is time to sync that branch up.

Yes, I was planning on updating that PR once you merge this in.

PR #109, opened 2025-01-08
Explore quantiseq tumor gene expression

### LINKED ISSUE #100: Explore tumor gene expression distribution for quantiseq inference
From #96, with bold added here for emphasis:

> One aspect I briefly peeked at in this notebook is the effect of including the argument tumor = TRUE, which removes a set of genes that are highly expressed in TCGA and is recommended when analyzing tumor samples. In our case, we are analyzing tumor samples, but pediatric tumor samples which have different biology from adult tumors. Turning this mode on led to more cell types being identified (again, only one sample right now!), but indeed some of those genes are have well above-average expression (as assessed quick and dirty with tpm) in the sample I looked at. **We would probably want to see more how these genes' expression levels look across more of our bulk samples before making a decision about using this argument.**

The `quanTIseq` method includes an option to consider tumor genes differently, and whether this is biologically relevant to our pediatric data is not clear. This issue tracks exploring the distributions of gene expression data that this argument affects; these genes are shown [in this list](https://github.com/icbi-lab/quanTIseq/blob/master/quantiseq/deconvolution/TIL10_TCGA_aberrant_immune_genes.txt). We can probably do something sort of quick-and-dirty here and just see where the TPM of these genes falls in the overall distribution with, for example, a faceted histogram for each project with one panel per library. 

I'll note that an alternative approach, since we don't have many samples, would be to just run quanTIseq with and without this option and directly compare results. The trouble with that is we don't have a sense of which is "correct," so it's not currently clear what criterion might be used to choose one set of results or the other. Hence, a quick look at gene expression values might provide some justification one way or the other!

sjspielman:
Closes #100 

`quanTIseq` has an option to exclude a subset of signature genes based on their aberrantly high expression in TCGA samples. While this option is recommended for tumor analysis, it's not immediately clear whether this applies to our _pediatric_ tumor samples.

This PR adds a notebook which does some exploration on the expression levels of those 17 so-called "tumor genes" compared to the other signature genes in our ScPCA samples. Specifically, I did one quick visualization, as well as some one-sided permutation tests to test if the tumor genes indeed have higher expression. In the end, I do not find strong evidence that these genes have higher expression, which suggests we don't need to use `tumor = TRUE` when running `quanTIseq`.

As part of this notebook, I also discovered that the gene symbols uses in the `quanTIseq` signature are not entirely up-to-date with what we expect for "official" gene symbols, and I made sure they were fully considered in this analysis. This will also be relevant to next steps where we actually deconvolute.

Here's the HTML:
[quantiseq-tumor-genes.nb.html.zip](https://github.com/user-attachments/files/18350426/quantiseq-tumor-genes.nb.html.zip)


analysis/bulk-deconvolution/exploratory-notebooks/quantiseq-tumor-genes.Rmd
```diff
@@ -0,0 +1,264 @@
...
+  purrr::map(
+    \(x) {
+      # read matrix and convert to table
+      readr::read_rds(x) |>
+        as.data.frame() |>
+        tibble::rownames_to_column(var = "gene_symbol")
+      }
+  ) |> 
+  # Combine into a single data frame
+  purrr::reduce(dplyr::left_join, by = "gene_symbol") |>
```
jashapiro, line 162: Just say I would probably have done this with `cbind` as we know the gene lists are the same... joining is slow!
```suggestion
  purrr::map(readr::read_rds) |> 
  # combine into a single matrix
  purrr::reduce(cbind) |>
  as.data.frame() |>
  tibble::rownames_to_column(var = "gene_symbol") |>
```

analysis/bulk-deconvolution/exploratory-notebooks/quantiseq-tumor-genes.Rmd
```diff
@@ -0,0 +1,264 @@
...
+    axis.text = element_text(size = rel(0.8)), 
+    axis.title = element_text(size = rel(0.8)), 
+    legend.position = "none"
+  )
+```
+
+Looking over a great many jitter plots, there does not seem to be a universal trend of abberantly high expression among tumor genes.
+
+
+## Permutation tests
```
jashapiro, line 202: Okay, but why not just do the straight tests? I would expect the non-parametric test to work just fine in this situation (and probably the t test too, since it tends to be pretty robust).

analysis/bulk-deconvolution/exploratory-notebooks/quantiseq-tumor-genes.Rmd
```diff
@@ -0,0 +1,264 @@
...
+      glanced <- purrr::map_df(models, broom::glance, .id = "id")
+      mod <- t.test(tpm ~ tumor_gene, alternative = "less", data = sample_tpm)
+      ttest_p <- mean(glanced$statistic > broom::glance(mod)$statistic)
+      
+      data.frame(
+        sample_id = x, 
+        nonparametric = wilcox_p, 
+        parametric = ttest_p
+      )
+    }
```
jashapiro, line 238: Expanding on my earlier comment, just for curiosity, what would happen if you just did this (just the straight tests, no permutations).
```suggestion
  purrr::map(
    \(x) {
      sample_tpm <- dplyr::filter(tpm_df, sample_id == x)
      
      wilcox_p <- wilcox.test(tpm ~ tumor_gene, alternative = "less", data = sample_tpm)$p.value
      ttest_p <- t.test(tpm ~ tumor_gene, alternative = "less", data = sample_tpm)$p.value
      
      data.frame(
        sample_id = x, 
        nonparametric = wilcox_p, 
        parametric = ttest_p
      )
    }
```

jashapiro (dismissed):
As this is exploration, this seems fine, but I had questions about why permutations here. Seems like a lot of work...

analysis/bulk-deconvolution/exploratory-notebooks/quantiseq-tumor-genes.Rmd
sjspielman: These tests were brought to you by my rough estimation that about 10% of the meetings I go to these days feature at least a brief discussion of why more people need to be shuffling data 😂  
So, I figured, perhaps today is my turn to shuffle data!

analysis/bulk-deconvolution/exploratory-notebooks/quantiseq-tumor-genes.Rmd
sjspielman: Hmm...well.... 
Screenshot 2025-01-08 at 1 57 33 PM


I made sure the directionality was right, too:

jashapiro: Okay, so that makes me a bit concerned that the permutations are not quite right. To be honest, I was not sure I followed the logic of your permutation code.

Looking back at it now, I think you actually have the directionality of the statistic comparisons backward. You should be looking for the observed statistic values to be _greater_ than the permutations.

sjspielman: > You should be looking for the observed statistic values to be greater than the permutations.

Ahhhhaaa that is the bug, glad we caught this, as I'm more inclined to think we should turn tumor mode on! I would tend to prefer to pick one setting and use it throughout all samples, but do you want to convince me otherwise?

sjspielman:
Somewhat bigger (though short code!) update - I toggled out the hypothesis tests for a mixed effects model, since I realized `lme4` (an old friend of mine...) was in our `renv` environment. I built a model controlling for sample and gene, and recovered a significant overall effect. I trust this more than a visual inspection of tiny jitter plots, so I updated the conclusion to be open to `tumor=TRUE`, which I will build in as an argument defaulting to TRUE over in #110.

[quantiseq-tumor-genes.nb.html.zip](https://github.com/user-attachments/files/18352594/quantiseq-tumor-genes.nb.html.zip)

jashapiro (approved):
Well, I'm not sure that is the most appropriate model, given the distributions of expression across genes, but I don't think it is worth getting out a glm or stan, and I agree with your conclusion that you should use `tumor=TRUE` for the analysis.

sjspielman:
> Well, I'm not sure that is the most appropriate model, given the distributions of expression across genes, but I don't think it is worth getting out a glm or stan, 

Oh yes it certainly should be more carefully done if we _really_ wanted the answers here! I think it's worth adding that caveat into the notebook, and stating plainly that we just want a quick sense for whether to use this argument.

PR #113, opened 2025-01-09
Regenerate manuscript figures

### LINKED ISSUE #107: Regenerate figures with updated dependencies

sjspielman:
Closes #107 

This PR updates all code-generated figures _except_ for those being addressed in #90 (#106). I generated them by running the `generate-figures-tables.sh` script. The apparent changes do not seem unreasonable given some of the changes to the `scpca-nf` pipeline since we last generated the figures. 

There is one code change I implemented - the panel box and axis lines seemed really thick so I dropped them to their defaults here. Is that ok?


allyhawkins (approved):
LGTM, note that we will also want to make sure these updated figure panels get updated in the compiled figures as well.

Issue #114, opened 2025-01-09
Update compiled figures


sjspielman:
> LGTM, note that we will also want to make sure these updated figure panels get updated in the compiled figures as well.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpca-paper-figures/pull/113#pullrequestreview-2540828622_
            
--- 
Following #107 & #90, we'll want to update compiled figures as well:

- [ ] Figure 1
- [x] Figure 2
- [x] Figure 3 (unchanged, free checkbox)
- [x] Figure 4
- [x] Figure S1
- [x] Figure S2
- [x] ~Figure S3~ This will be accomplished as part of #91, so we get a free check here
- [x] Figure S4
- [x] Figure S5
- [x] Figure S6
- [x] Figure S7

PR #115, opened 2025-01-10
Recompile figures, PR 1/~3

sjspielman:
Towards #114

This PR updates compiled figures 2, 4, S1, and S2. 

I also did some small theme tweaking along the way to reduce the amount of manual modification we need to do in Illustrator, which is why there are some changes to individual panels. I also fixed the patchwork code for S1A to get the exported panel to also have the legend on the top, so it doesn't have to be moved in Illustrator. The updated panels match what is included in each compiled figure.


figures/compiled_figures/pngs/figure_s2.png
allyhawkins, line 1: Just noting that the size of the text on the axis labels is different for B, C, and D between these two. I think we want to match the original font size.

sjspielman: Oh I did want to say something else about this figure which I forgot!!

Updated this figure confused me... the compiled PDF had individual vector graphics panels here, but we only export a single PNG image to place in. Looking a bit at the history of this file, it was indeed PDF at one point but then that got reverted: https://github.com/AlexsLemonade/scpca-paper-figures/commits/3e5d2d701447576be6f4a2c008eb93636c29a66f/figures/pngs/FigS2B_adt-plots.png?browsing_rename_history=true&new_path=figures/pngs/FigS2B-D_adt-plots.png&original_branch=main

I don't exactly understand what originally went down here to end up with a PNG in the repo but panels of PDFs in the compiled version. I filed this PR though, so, if anyone were to understand it would be me 😬 

In any case, maybe we should take the time now to think more about how to treat this figure! Maybe I should re-PDF-ify it and remind ourselves of whether a size restriction was the problem? If it's a PDF, we can always modify size in Illustrator directly. But if it's a PNG, then we can only do it via code.

allyhawkins: I feel like this might help us here: https://github.com/AlexsLemonade/scpca-paper-figures/pull/55
I do think size was an issue, but it looks like we made everything into a pdf except for panel c?

sjspielman: Good find, thanks. I will update the script to actually export plots so that b & d are pdf, and c is png, and will recompile.

sjspielman: > I don't exactly understand what originally went down here to end up with a PNG in the repo but panels of PDFs in the compiled version. I filed this PR though, so, if anyone were to understand it would be me 😬

Also I now understand what was happening here - they actually were PNGs, but the _text_ had been written over, which made it look like the object was from a PDF even though it wasn't.

allyhawkins: > Also I now understand what was happening here - they actually were PNGs, but the text had been written over, which made it look like the object was from a PDF even though it wasn't.

I think maybe the text had been written over in an attempt to get the correct font size in the combined figure?

sjspielman: > I think maybe the text had been written over in an attempt to get the correct font size in the combined figure?

💯 that's the reason! Trying to get it into the script though so we don't have to keep doing that, since it's not pleasant. Stay tuned after pre-planning :)

sjspielman:
I think Figure S2 looks reasonably consistent now. For some reason Illustrator isn't exporting the full artboard with the PNG even though it's not selected, but seems fine enough.

allyhawkins (approved):
LGTM

Issue #117, opened 2025-01-13
Add notebook comparing EPIC and quanTIseq


sjspielman:
Following #116, we will want to add notebook that explore the EPIC results and compare them to quanTIseq. 

Part of this comparison, at a minimum, should entail what proportion of "cells" in each dataset were classified vs. remain other/unknown. We'll want to compare this among results for both EPIC references and quanTIseq. 

It might also be helpful to bring some of the quanTIseq exploration we performed in the [quanTIseq notebook](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/bulk-deconvolution/exploratory-notebooks/explore-quantiseq-results.Rmd) into this one so we can have a clear comparison between methods. Edit: In particular, we'll want a stacked barplot of cell types _including_ "other"; this is not precisely in the other notebook (we have a [stacked barplot without other](https://github.com/AlexsLemonade/scpca-paper-figures/blob/0c9a5592dcc48aed2a21cd2fbe9be3cd55ed7bc5/analysis/bulk-deconvolution/exploratory-notebooks/explore-quantiseq-results.Rmd#L201-L203)), but we'll do it this way here.

sjspielman:
I'm beginning to write up the second notebook comparing EPIC and quanTIseq inferences, and I'm not exactly sure what the right way is to compare cell type proportions across references/methods. Since each reference has a different set of cell types, this will influence the values themselves. I had originally thought maybe it makes sense to renormalize each dataset without "other," but since "other" isn't directly comparable between references, I'm not convinced this changes the situation. All the ways I can think to compare feel too much like "apples to oranges" and not very robust since they are each conditioned on different cell type distributions.

Any thoughts on _quantitatively_ comparing individual fractions @allyhawkins? I'm starting to think that maybe we can really only do this qualitatively, and this next notebook would more serve to see all results in one spot. But then again if that's all we can do, maybe it actually suffices to export select PNGs from the individual method exploration notebooks?

allyhawkins:
I think you could still compare the "other" fraction and state the caveat that the you may be missing some cell types that are not available in the reference for a specific method. 

There are some cell types that are the same across methods (macrophages, B cells, T cells, Monocytes, etc). I would directly compare the proportions that get classified as each of the cell types that overlap. 

[...]

PR #120, opened 2025-01-15
Notebook to explore EPIC results

sjspielman:
Towards #117 

This PR adds a notebook which explores the EPIC results and chooses a reference to move forward with. Spoiler, I think `TRef` is probably the winner (with caveats that I mention in the notebook!). Originally, I had planned to do a single notebook that explores EPIC and compares to quanTIseq, but as I was writing this up I felt this was probably long enough for one notebook, and we should do a separate notebook aimed more specifically at comparison.

Much of this code is similar to what I did to explore the quanTIseq results, but not all of it. I think this should be mostly self-explanatory for review (it's mostly exploratory plots!), but please let me know what I can further clarify! Here's the HTML:
[explore-epic-results.nb.html.zip](https://github.com/user-attachments/files/18430127/explore-epic-results.nb.html.zip)


analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+  facetvar = project_diagnosis,
+  fillvar = forcats::fct_relevel(epic_celltype, celltype_levels),
+  plot_title = "BRef reference", 
+  color_map = color_map
+) 
+
+tref_barplot / bref_barplot
+```
+
+Generally speaking, `TRef` identifies more cell types compared to `BRef`, although whether it _should_ be classifying all those cells is an open question to bear in mind!
```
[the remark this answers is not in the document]
allyhawkins: Also I think it's important to note that many of the osteosarcoma samples have a majority of cells being assigned to CAFs. I'm curious what genes are associated with the CAF cells? There's probably a lot of incorrect assignment there since CAFs tend to look very similar to tumor cells and even more similar to osteosarcoma if things like collagen and ECM genes are in that gene list.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+
+shared_celltypes |>
+  purrr::map(compare_reference_celltype, epic_df) |>
+  patchwork::wrap_plots(nrow = length(shared_celltypes), guides = "collect")
+```
+From all above comparisons, we see these trends:
+
+* `Bcell`, `CD4_Tcells`,  and `NKcell` proportions are roughly similar between references and low across all samples
+  * In a handful of CNS tumor samples, `TRef` estimates higher `BCell` proportions
+  * In a handful of CNS tumor samples, `TRef` estimates higher `BCell` proportions
```
allyhawkins, line 241:
```suggestion
  * In a handful of CNS tumor samples, `TRef` estimates higher `BCell` proportions
```

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+
+Generally speaking, `TRef` identifies more cell types compared to `BRef`, although whether it _should_ be classifying all those cells is an open question to bear in mind!
+
+Let's compare the proportions of shared cell types inferred between references a bit closely with scatterplots:
+
+```{r fig.width = 12, fig.height = 14}
+shared_celltypes <- c("Bcells", "CD4_Tcells", "CD8_Tcells", "NKcells", "otherCells")
+
+
+compare_reference_celltype <- function(celltype, epic_df) {
```
allyhawkins, line 210: I don't love these plots... I think I would rather just see a sina plot where you show the proportions of each cell type on the y-axis and the reference on the x-axis.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+  * In a handful of CNS tumor samples, `TRef` estimates higher `BCell` proportions
+  * In a handful of CNS tumor samples, `TRef` estimates higher `BCell` proportions
+  * It does look like `CD4_Tcells` estimates are a bit higher for `TRef`, but there are two osteosarcoma samples which show really high values from `BRef`
+* `CD8_Tcells` are mostly similar with some additional trends:
+  * Vaues are generally higher for `BRef` for Wilms Tumor samples
+  * Values are generally higher for `TRef` and both CNS tumor and osteosarcoma samples
+* As stated, `TRef` classifies more cells than does `BRef` for all but one osteosarcoma sample
+
+
+**Based on these comparisons, we'll move forward with the `TRef` reference.**
```
allyhawkins, line 249: I don't disagree that Tref is the better option, but I think that's more because of the cell types that Tref has that Bref doesn't have NOT because Tref identifies a higher proportion of cells for the shared cell types. I think it's probably more important to think about the biology of each of these tumors and what normal cell types we may expect to see.

allyhawkins: In addition to the shared phenotypes, I would look at the monocyte and macrophage fraction in each sample. My guess is that there's a correlation in samples that have monocytes in Bref and macrophages in Tref. Although the macrophage population is probably consistently smaller, I would expect there to be correlation. This may not be important at this point to dive into, but I think that's an important thing to note.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+
+```{r}
+tref_df <- epic_df |>
+  dplyr::filter(reference == "TRef")
+```
+
+
+Let's look at the samples had very few "other" cells:
```
allyhawkins, line 257:
```suggestion
Let's look at the samples had very few "other" cells (those with fraction < 0.1):
```
Specifying in the text what we mean by "few"

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+    glue::glue("{sample_id}\n({diagnosis})"), 
+    sample_id)
+  ) |>
+  dplyr::select(
+    project_diagnosis, 
+    sample_diagnosis, 
+    epic_celltype, 
+    fraction) 
+
+ggplot(classified_samples_df) +
```
allyhawkins, line 278: I don't think you need this plot at all if you already have the above plot and just blow it up a little. I would instead just add a summary to that plot that includes many of the bullet points listed here.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+    fraction) 
+
+ggplot(classified_samples_df) + 
+  aes(x = sample_diagnosis, y = fraction, fill = epic_celltype) +
+  geom_col() + 
+  scale_fill_manual(values = color_map, na.value = "grey90", name = "cell type") + 
+  theme(axis.text.x = element_text(angle = 30, hjust = 1)) + 
+  facet_wrap(vars(project_diagnosis), scales = "free_x")
+```
+* It does make sense that `CAFs` are dominating osteosarcoma samples (indeed they are prevalent, generally speaking, across a few more than these osteosarcoma samples) since we know they are a major component of osteosarcomas
```
allyhawkins, line 285: I think that a very very large proportion of these assignments are incorrect. CAFs and osteo cells are going to be hard to classify because I would assume the genes being used to define a "CAF" is basically the same as what defines a sarcoma cell. To define CAFs vs. tumor here we need to be more specific than the Tref reference probably is. That being said, I think the Bref actually does a better job for these samples than Tref. I would expect to see monocytes in osteo samples and with Bref you actually are identifying "other", which we assume are tumor cells. If we move forward with EPIC I don't think we need to choose one reference for all projects. The cell types in a glioma will not be the same as cell types in a sarcoma.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+# convert number to proportion
+samples_per_celltype_df <- epic_df |>
+  dplyr::filter(is_other) |>
+  dplyr::count(project_diagnosis) |>
+  dplyr::right_join(samples_per_celltype_df) |>
+  dplyr::mutate(prop_samples = n_samples / n)
+```
+
+```{r fig.height=4, fig.width=12, message = FALSE}
+ggplot(samples_per_celltype_df) +
```
allyhawkins, line 349: Is this plot supposed to have both Tref and Bref in the same plot? Cell types from both are listed and I think I would actually separate these out so you show Tref and Bref separately. I also would sort by the proportion.

allyhawkins: I think that's why you see some numbers that are wrong. For example, CAFs are in almost all samples but that is not represented by this plot.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+    cols = vars(project_diagnosis)
+  )
+```
+Overall values seem highly similar, but we can see that `Macrophages` have been corrected to have slightly lower values.
+A more in-depth comparison here might zoom in to axes and/or explore relative error between measurements, but that is beyond the scope of what we are interested in looking at which is just to peek and get a sense!
+
+## Conclusions
+
+Between the two `EPIC` references explored here, `TRef` seems more suited to our data since it contains some non-leukocyte normal cell types and does a more thorough job of classification.
+That said, it is not clear whether it _should_ do a more thorough job of classification, and it is missing useful-to-have cell types `Monocytes` and `Neutrophils` which are in the `BRef` reference.
```
allyhawkins, line 392: Exactly! Just because it identifies cells doesn't mean it should, looking at you Osteo 👀

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+- The `BRef` reference (6 + other):
+  - **`Bcells`**   
+  - **`CD4_Tcells`**
+  - **`CD8_Tcells`**
+  - `Monocytes`
+  - `Neutrophils`
+  - **`NKcells`**
+  - `otherCells`
+          
+Notably, `TRef` includes fibroblasts and epithelial cells, which we are interested in being able to quantify!
```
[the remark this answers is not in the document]
allyhawkins: Another thing is that these are _cancer_ associated fibroblasts, not fibroblasts. There is a difference and a lot of controversy about what is a CAF since the features of CAFs can be tumor dependent.

allyhawkins (commented):
Most of this looks good, I just have some very strong opinions about classifying osteo cells as CAFs, which I left throughout. I also generally think we should focus more on whether or not the cell types fit with expected biology rather than exact numbers and proportions. Those are important too, but not enough to pick one reference over another. 

That being said, if we move forward with EPIC I would want to use different references for each project. Tref makes sense for the gliomas and CNS tumors, but Bref makes sense for osteo. Wilms who knows... I don't know enough about that tumor but I think a key question is whether or not we expect to see B cells in those tumors? If so, then I would use Tref, if not, then I would use Bref. 

Also, I agree about doing the comparison to quantiseq separately.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,401 @@
...
+
+
+```{r fig.height=8, fig.width=16}
+# First, we'll turn the otherCells level into `NA` for ease in coloring
+epic_otherna_df <- epic_df |>
+  dplyr::mutate(epic_celltype = ifelse(
+    epic_celltype == "otherCells", 
+    NA_character_, 
+    epic_celltype
+  ))
```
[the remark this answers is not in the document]
sjspielman: Ah you know, the only reason I did this is because originally I was using a pre-defined palette, and this was the easiest way to get these cells coming up gray using `na_color`! I'll go ahead and properly add them to the palette vector.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
sjspielman: > This may not be important at this point to dive into, but I think that's an important thing to note.

This is an interesting question. Agree, this is not the time for a deep dive, but certainly worth a plot.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
sjspielman: oh 💯

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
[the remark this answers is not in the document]
sjspielman: > There is a pretty big difference!

omg I know, and I seem to do this all the time. not a great typo.. thank you!!!

> That being said, if we move forward with EPIC I would want to use different references for each project. Tref makes sense for the gliomas and CNS tumors, but Bref makes sense for osteo. Wilms who knows... I don't know enough about that tumor but I think a key question is whether or not we expect to see B cells in those tumors? If so, then I would use Tref, if not, then I would use Bref.

This didn't even occur to me, but it makes much more sense! I'll poke around at the literature a bit for this.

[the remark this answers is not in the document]
sjspielman: - Based on this reference about WT microenvironment, B cells are indeed present: https://www.sciencedirect.com/science/article/pii/S1477513119300695#fig2. 
- But here, it describes that `Macrophages` are quite prevalent, but it doesn't say anything about B cells (absence of evidence is notably not evidence of absence though!); scroll to section `Immune cell infiltration` in https://www.cell.com/molecular-therapy-family/oncology/fulltext/S2372-7705(21)00098-X
- I went back and looked at the quanTIseq results whose reference has both macrophages and B-cells, and B-cells were inferred in all but one sample, and macrophages were only inferred in a subset (although sometimes with high proportions!). 

Taken together, I'm leaning towards BRef. Any thoughts, @allyhawkins ?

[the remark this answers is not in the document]
allyhawkins: I also was looking at the results again for each of the refs and it looks like Trefs is calling a lot of CAFs, which for wilms are probably tumor cells. Because of that I think I would use Bref.

sjspielman:
This should now be ready for another look! I've overhauled the notebook to not focus on "choosing a reference," and instead open the notebook with a statement that "the references are selected based on biological expectations," and mostly explore the results we have. I think these references make more sense, but I'm on the fence about EPIC itself, which we'll disentangle in the next notebook comparing to quanTIseq! This notebook gives us an initial sense of what we might expect in that next notebook.

While working on this, I revisited the definition of the mRNA proportions which I don't think I had fully understood before - but I get it now, so I updated the README with the correct description! I don't think we really need to look at these values, though.

Here's the current version of the notebook:
[explore-epic-results.nb.html.zip](https://github.com/user-attachments/files/18459006/explore-epic-results.nb.html.zip)

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,288 @@
...
+
+
+## Visualization
+
+### Proportions of cell types
+
+Considering all categories of cells, what proportions of cell types are represented across samples?
+
+
+```{r fig.height=8, fig.width=12}
```
allyhawkins, line 128: Thank you for making these panels bigger! But I actually think we want to show the results for both references for each reference here? I think that drives home the point of why you are choosing one ref over the other in many cases.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,288 @@
...
+In most samples, the vast majority of cells are `otherCells`, although there are two samples which are estimated to have ~100% cells classified; it is very likely that cells in these samples (and possibly more samples) are incorrectly classified.
+
+We also see these trends:
+
+* For the `TRef` projects:
+  * Macrophages are most commonly observed in `SCPCP000001` samples
+  * Endotheliel cells tend to dominate `SCPCP000001` and `SCPCP000002`
+* For the `BRef` projects:
+  * CD4 T-cells and monocytes are more prevalent in `SCPCP000017` compared to `SCPCP000006`
+  * CD8 T-cells dominate `SCPCP000006` cell types
```
allyhawkins, line 172: I might just add a quick comment that this lines up with expectations, at least the macrophages and endothelial in brain/cns and the monocytes in osteo.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,288 @@
...
+
+Let's have a quick look at the proportion of classified cells on its own:
+
+```{r}
+epic_classified_df <- epic_df |>
+  dplyr::filter(epic_celltype == "otherCells") |>
+  dplyr::mutate(fraction_classified = 1 - fraction) 
+
+ggplot(epic_classified_df) +
```
allyhawkins, line 181: Can you make these sina plots instead of box plots? 
I also would add a comment that this is looking specifically at cells that are not categorized as "other", which we assume to be tumor cells.

allyhawkins: I wouldn't expect more than ~20-30% of cells be classified, so these results do line up with that which is good!

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
```diff
@@ -0,0 +1,288 @@
...
+  theme(axis.text.x = element_text(angle = 30, hjust = 1, size = rel(0.9)))
+
+```
+In the vast majority of samples, fewer than 25% of cells are classified, with the fewest cells classified in Wilms Tumor. 
+
+
+Notably, across samples, there appear to be very few B-cells inferred, and when they are inferred the frations are extremely small (to the point that we might essentially consider them to be 0).
+Let's look at this a bit more closely, zooming in on the section of the plot with most of the data:
+
+```{r, fig.width = 10}
```
allyhawkins, line 193: What's the reasoning behind zooming in on B cells here? I don't think you need it.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
allyhawkins: After going through the rest of the notebook, I would show this plot for both references for all projects and then include reasoning on why you are picking each reference for each project. And then for the rest of the notebook you could just use the one you pick.

analysis/bulk-deconvolution/exploratory-notebooks/explore-epic-results.Rmd
allyhawkins: This plot still has some issues. Looking at the stacked bar chart, endothelial cells should be in 100% of samples from SCPCP000001 and SCPCP000002, but the proportion is not even .15?

allyhawkins (commented):
I think the new updates look good! The biggest thing is that I still want to see the stacked bar chart for both references for each project. That plot is pretty important and helps drive home why you would choose one reference over another.

sjspielman:
Next version! I believe I've hit all the reviews, let me know if anything else would be helpful!
[explore-epic-results.nb.html.zip](https://github.com/user-attachments/files/18495443/explore-epic-results.nb.html.zip)


I did have the thought while I was doing this that it might be nice to also explore which labels were applied to the same cells from the two references, but this seemed a bit overkill for _this_ notebook. Instead, I plan to include that kind of analysis in the next notebook comparing epic and quantiseq.

allyhawkins (approved):
LGTM!

sjspielman:
> I did have the thought while I was doing this that it might be nice to also explore which labels were applied to the same cells from the two references, but this seemed a bit overkill for this notebook. I

For posterity, I will note that when I wrote this my brain apparently temporarily forgot that this is extremely not actually single-cell data! There will be no comparing at a cell-by-cell level, since that just doesn't make sense in the first place here in _bulk deconvolution_ land.

PR #121, opened 2025-01-15
Update images in Figure 3 and recompile

### LINKED ISSUE #60: Use higher quality images for 3A and 3B
👍🏻  One thing we might want to keep track of/an eye on is if the PNGs underlying panels A and B should be saved at a higher quality in the future. My PDF to PDF conversion senses are tingling. But I don't think that needs to hold things up.

_Originally posted by @jaclyn-taroni in https://github.com/AlexsLemonade/scpca-paper-figures/pull/59#pullrequestreview-1880485732_

### LINKED ISSUE #91: Update download images for Fig. 3A and 3B to reflect new download folder structure
The file names in the downloads have slightly changed since we created figure 3. We should first update the images in this repo to be the most current images in our docs (see [project download image](https://github.com/AlexsLemonade/scpca-docs/blob/main/docs/images/project-download-folder.png) and [merged download image](https://github.com/AlexsLemonade/scpca-docs/blob/main/docs/images/merged-project-download-folder.png)). Then we need to update the compiled version of Figure 3. 

We may also want to tackle #60 at the same time.

allyhawkins:
Closes #60 
Closes #91 

Here I'm adding higher quality images for the download structure shown in Fig 3A and 3B. We decided to use SVGs instead of PNGs so I added those and removed the old ones. I also removed an old PNG that we weren't actually using in the figures. I updated the compiled figure to have these new images and the updated Fig 3D merged UMAP that was updated when the `renv` updates occurred.


sjspielman (commented):
This looks fine to me, but I do worry that the new figure panel font is too small.  But, I don't think this can be changed without reducing the amount of text in the "header" which can't really be done I think. Do you have any thoughts on font size?

allyhawkins:
> This looks fine to me, but I do worry that the new figure panel font is too small. But, I don't think this can be changed without reducing the amount of text in the "header" which can't really be done I think. Do you have any thoughts on font size?

I think it's legible and similar to the font size used in the other panels even if it is smaller than what it was before. I agree that we can't really change the header so we're kind of stuck with this new size unless we totally reorganize this figure... but I'm not sure how that would work to fit everything properly and still get big enough font size.

allyhawkins:
> > This looks fine to me, but I do worry that the new figure panel font is too small. But, I don't think this can be changed without reducing the amount of text in the "header" which can't really be done I think. Do you have any thoughts on font size?
> 
> I think it's legible and similar to the font size used in the other panels even if it is smaller than what it was before. I agree that we can't really change the header so we're kind of stuck with this new size unless we totally reorganize this figure... but I'm not sure how that would work to fit everything properly and still get big enough font size.

I think the file types in the little icons are a little hard to read. We could remove those since the file extensions are still listed? Other than that, I'm not sure how else we could change it without removing content.

sjspielman (approved):
I think the answer is it's fine for now, and we can circle back if/when the team has comments when we all review the next draft version!

Issue #97, opened 2024-12-17
Test deconvolution methods: EPIC


sjspielman:
_Broken out of #126_

This issue tracks exploring usage of `EPIC` for bulk cell type deconvolution. Sources:
- Paper: <https://doi.org/10.7554/eLife.26476.001>
- Package: <https://github.com/GfellerLab/EPIC/>

PR #122, opened 2025-01-22
Add epic/quantiseq comparison notebook

### LINKED ISSUE #117: Add notebook comparing EPIC and quanTIseq

sjspielman:
Closes #117 

This PR adds a notebook to compare EPIC and quanTIseq results. I'm opening this PR as a Draft to indicate that the notebook isn't quite done yet - I still need to add some text interpreting the plots! - but I wanted get any thoughts on additional/alternative viz. Again, I'm not confident we can do a rigorous quantitative comparison here, as I lay out in the opening caveats in the notebook, so this is really just plots. 

Here's a rundown of what's here plot-wise:
- Barplots across samples, of all inferences shown with and without "other"
- Barplots across samples, of proportions of cell types the references have in common. I'll note that these are kind of hard to read, but also I don't think you need to read them too carefully - the punchline is going to be "huh, these are really different!" and that message is there even without the finer interpretation details!
  - I originally made these as scatterplots, but changed to barplots to show a different strategy. But, the scatterplot at the end of the document showing classified proportion gives the vibe for what scatterplots here would look like!
- two plots (sina which we should def keep, scatterplot which could go either way) looking just at the classified proportion

For this initial review, please let me know any high-level comments about plotting strategies to be tweaked or re-considered, and then I will make those updates and flesh out all the interpretation text for and request a "real" aka thorough review. You can feel free to go for a more nitty gritty review though if you think we're ready for that!

Here's the current version of the notebook
[compare-epic-quantiseq.nb.html.zip](https://github.com/user-attachments/files/18507700/compare-epic-quantiseq.nb.html.zip)


analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
```diff
@@ -0,0 +1,422 @@
...
+classified_prop_df <- consensus_df |>
+  dplyr::filter(
+    !(sample_id %in% fully_classified_df$sample_id), 
+    celltype_consensus == "Other"
+  ) |>
+  dplyr::mutate(
+    classified_proportion = 1 - proportion
+  )
+
+ggplot(classified_prop_df) +
```
allyhawkins, line 382: I think it's worth pointing out a few things about why we might be seeing differences or similarities in the proportion of cells classified as "other". (I'm also writing this mostly for myself to help figure out what's going in these results). 

- For 01 and 02, these are mostly the same but there's NK cells being identified in quantiseq that perhaps are not identified in Epic for whatever reason? But also, we get a lot of endothelial cells with epic that we don't see with quantiseq. I'm not totally sure which one is correct, but I think that's worth pointing out. 
- With 06 we see that monocytes and CD8 T cells are identified in both but there are more cell types identified with quantiseq. 
- For 09:CNS we also see endothelial cells in epic, but NK cells in quantiseq. 
- Osteo looks pretty similar between both methods, but you are getting more granular than just monocytes with quantiseq.

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
```diff
@@ -0,0 +1,422 @@
...
+  "Macrophages.M2" = "sienna1",
+  "Dendritic.cells" = "purple",
+  "Tregs" = "blue"
+) 
+```
+
+
+And now to the plots:
+
+```{r, fig.width = 12, fig.height = 7, warning = FALSE}
```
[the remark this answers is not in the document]
allyhawkins: Another note is that for 01, 02, and 09 we should combine the color for M1, M2, and macrophages so that there is only one macrophage in the legend. Right now it's really hard to distinguish M1 and macrophages.

allyhawkins (commented):
Generally this looks pretty good. My main comment is that I don't love the grouped bar plots to look at the shared cell types and instead commented with an idea about how to do a stacked bar chart for that. 

I think the biggest difference between the two methods is that quantiseq has more granular immune cell types (especially on the monocyte -> macrophage lineage), but it is missing endothelial cells. I'm not sure I really care about CAFs being missing since those are going to be harder to distinguish from cancer cells so I'm generally iffy about their presence. I wonder how much truth there is in the endothelial cell population since I expect being able to identify those vs. immune cell types with gene expression should be relatively easy. So the fact that they are present with epic makes me wonder if they're real. But then again we also see that NK cells are mostly absent from epic but present in quantiseq... 

I'm curious how the references are built for each of these methods? Is it a list of marker genes for each of the cell types? Is there a way to use "custom" references?

sjspielman:
> I'm curious how the references are built for each of these methods? Is it a list of marker genes for each of the cell types? Is there a way to use "custom" references?

I assume the spirit of this question is, "can we build our own reference?", and if so, yes I am thinking about this as well! I'll note that `quanTIseq` doesn't currently accept a custom reference (one could fork their repo and probably modify that, but not worth the trouble imo). `EPIC` references consist of three (one optional) components:
- a "reference profile" which is a gene-by-cell-type expression matrix (can be raw counts, TPM, FPKM, but NOT normalized. ideally it's TPM). This profile should also be well-larger than the signature gene set; it's meant be all gene expression in the cell types of interest. These constraints mean we can't use `celldex` (all of them are logcounts), `CIBERSORT`'s signature matrix, and or `quanTIseq`'s signature matrix which I already tried and it was a little bonkers https://github.com/AlexsLemonade/scpca-paper-figures/issues/97. The reason the latter two don't work is that they are limited to signature genes specifically used for discriminating signature matrices, but the genes in EPIC reference profiles are not signature genes. EPIC is looking for full profiles for the cell types.
- optionally, the variability of each gene in each cell type based on the datasets the reference was compiled from
- a subset of the full reference profile, a set of signature genes we expect to be highly expressed in at least some of the non-tumor cells, but not in the tumor cells. these are not cell-type specific marker genes.

In the EPIC paper, they built their two references from different source data sets and calculated values as:
> The median value of TPM counts was computed per cell type and per gene. Similarly, the interquartile range of the TPM counts was computed per cell type and gene, as a measure of the variability of each gene expression in each cell type.

So, it might be a bit of a lift to make our own reference that fits these assumptions, but it's worth discussing for next sprint I think. For now I am also playing around a bit with MCP Counter (you get scores, not fractions here) which - spoiler - appears to think samples are generally abounding with cancer-associated fibroblasts.

Another thought I want to put out there is, I think it's worth revisiting the initial gene name conversion I did here. I only checked our gene names against their signature list, but it's possible we have a ton of mismatches with the gene names in the reference profile itself that could be affecting results. Probably worth a notebook to check?

allyhawkins:
> I assume the spirit of this question is, "can we build our own reference?", and if so, yes I am thinking about this as well!

The reason I asked this is because yes I want to build our own reference, but also can we somehow combine the existing references for these methods. Mainly, could we combine Bref and Tref? But I don't know that its worth the effort. I would love to use the same cell types available in `celldex`, but also probably a lot of effort. I think it could be done, we just couldn't use `celldex`, we would have to go straight to the Blueprint and ENCODE data. From the celldex docs for `BlueprintEncodeData()`.  

> This function provides normalized expression values for 259 bulk RNA-seq samples generated by
Blueprint and ENCODE from pure populations of stroma and immune cells (Martens and Stunnenberg, 2013; The ENCODE Consortium, 2012). The samples were processed and normalized as
described in Aran, Looney and Liu et al. (2019), i.e., the raw RNA-seq counts were downloaded
from Blueprint and ENCODE in 2016 and normalized via edgeR (TPMs).

So I think if we were really focused on getting an exact comparison with the single-cell data we should do that, but I don't think it's necessary. I think I'll stick with my original comment about picking one to put in the main figures and one to put in the supplemental, so really this notebook is just showing us that there are some similarities (like overall "other" population) and some differences, like cells being classified as NK vs. endothelial cells. But these are all limitations of the references and things we should be sure to include discussion on. 

> Another thought I want to put out there is, I think it's worth revisiting the initial gene name conversion I did here. I only checked our gene names against their signature list, but it's possible we have a ton of mismatches with the gene names in the reference profile itself that could be affecting results. Probably worth a notebook to check?

If there's a way to easily do this then sure. Are all the genes in the reference profile used when doing the actual deconvolution or is it just the signature genes? If just the signature genes are used for deconvolution, then it's probably not worth it.

sjspielman:
> Mainly, could we combine Bref and Tref? 

I did look at this too earlier today. They were created from different underlying datasets - TRef is from immune cell profiles in general, and BRef is from tumor-infiltrating cells in melanoma samples. I'm not sure there's a clear way to combine them given their fundamentally different underlying biology. Plus, logistically, the values for the same cell types are fairly different - they are definitely correlated, but certainly not the same. There are also different numbers of genes and different gene signatures between these datasets, necessarily. 

> If there's a way to easily do this then sure. Are all the genes in the reference profile used when doing the actual deconvolution or is it just the signature genes? If just the signature genes are used for deconvolution, then it's probably not worth it.

No, the reference is really important - the signature genes should not be called signature genes since they are not cell type marker genes. All cell-type specific information is in the reference, so I think this has to be looked at.

sjspielman:
@allyhawkins FYI it seems we indeed have _quite_ a few ID mismatches - https://github.com/AlexsLemonade/scpca-paper-figures/issues/125#issuecomment-2609740506

I'm therefore going to put this PR on hold so we can re-run EPIC with a reference profile whose symbols better match our data!

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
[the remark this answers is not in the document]
sjspielman: I'm getting back to this today (now putting the ensembl-ified EPIC reference on pause 🎢 ), and want to clarify here:


> Another note is that for 01, 02, and 09 we should combine the color for M1, M2, and macrophages so that there is only one macrophage in the legend. Right now it's really hard to distinguish M1 and macrophages.

I'm not sure what exactly this is asking for.. I get that it's hard to distinguish M1 and Macro visually, which I kind of did on purpose! I wanted all the macrophage to be orange for easier visual comparisons, and further M1 and Macrophage never appear in the same reference. I can still change it for sure! What I'm not following is your suggestion to combine the color into 1 - are you suggesting that we collapse m1 & m2 in the first set of barplots that show all inferences from the methods, or in the second set of bar plots emphasizing shared cell types? I was assuming the latter, right?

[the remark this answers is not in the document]
allyhawkins: > are you suggesting that we collapse m1 & m2 in the first set of barplots that show all inferences from the methods, or in the second set of bar plots emphasizing shared cell types? I was assuming the latter, right?

Sorry, I realized after I posted this comment that for the purpose of the plot that is currently present, they will never appear in the same panel so it's all good the way you have it. That being said, we should combine them for the second set of bar plots that emphasizes the shared cell types.

sjspielman:
Next version of this notebook is here, now with some interpretive text!
[compare-epic-quantiseq.nb.html.zip](https://github.com/user-attachments/files/18541394/compare-epic-quantiseq.nb.html.zip)

Here's to hoping we can devise a reference of our own next...!

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
```diff
@@ -0,0 +1,441 @@
...
+
+
+### Conclusions
+
+
+* Across all inferences from both methods, B cells are fairly uncommon, but `quanTIseq` generally estimates higher fractions compared to `EPIC`
+* Considering projects `SCPCP000001` and `SCPCP000002` (high grade and low grade gliomas, respectively):
+	* NK cells are commonly identified by `quanTIseq` but not by `EPIC`
+	* There appears to be some amount of consistency between macrophage inferences between methods
+	* `EPIC` estimates substantial portions of endothelial cells, which are not present in `quanTIseq`. Similarly, `quanTIseq` estimates substantial proportions of dendritic cells, which are not present in `EPIC`.
```
allyhawkins, line 351: Maybe also add a note about Monoctyes. It looks like a large portion of these samples are either Monoctyes, Macrohpages, or NK cells.

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
```diff
@@ -0,0 +1,441 @@
...
+      
+    }
+  )
+```
+
+
+### Conclusions
+
+
+* Across all inferences from both methods, B cells are fairly uncommon, but `quanTIseq` generally estimates higher fractions compared to `EPIC`
```
allyhawkins, line 347: I think I might also note the overall differences in T cell populations. It looks like EPIC calls a lot of T cells and virtually no NK cells, while quanTIseq has a lot more NK cells and smaller proportions of T cells (probably due to the difference in granularity between immune cell types present in the different references).  

It looks like you mention it for some specific projects below, but I think the T cell differences also apply to the gliomas.

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
allyhawkins: Also just a side note that endothelial cells are not present in the top represented cell types using the consensus labels for this project, so that might be a reason at least for these projects to not use EPIC (at least with this reference).

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
```diff
@@ -0,0 +1,441 @@
...
+	* CD8 T cells as well as monocytes dominate the `EPIC` results, and while we don't see CD8 as commonly in `quanTIseq`, we do also see monocytes although often in _different_ samples from where this cell type is common according to `EPIC`
+	* `EPIC` results here have fewer cell types in general compared to `quanTIseq`, even considering the difference in references
+* For project `SCPCP000009` (cns tumors):
+	* We see similar patterns to `SCPCP000001` and `SCPCP000002` where NK cells are commonly identified by `quanTIseq` but not by `EPIC`,  `EPIC` estimates substantial portions of endothelial cells, and `quanTIseq` estimates substantial proportions of dendritic cells
+	* We also see T cells, in particular CD4, dominating in `EPIC` but not in `quanTIseq`
+	* In the sample where `EPIC` classifies nearly 100% of the cells, there are similar cell types between methods but vastly different proportions
+* For project `SCPCP000017` (osteo):
+	* Monocytes and CD4 T cells dominate in `EPIC` but these are much less common in `quanTIseq
+	* In the sample where `EPIC` classifies nearly 100% of the cells, the cell types dominating in `EPIC` tend not to be those dominating in `quanTIseq`.
```
allyhawkins, line 361: I think this is a good observation! We don't expect all cells to be normal, so the fact that we see this in EPIC but not quanTIseq might also be an argument for quanTIseq.

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
```diff
@@ -0,0 +1,441 @@
...
+  ) + 
+  scale_color_brewer(palette = "Dark2") +
+  theme(axis.text.x = element_text(angle = 30, hjust = 1))
+```
+
+Distributions of classified proportions, across all projects, are generally below 0.4 with a handful of outliers. 
+As noted in the barplots, `EPIC` tends to classify lower proportions of each sample compared to `quanTIseq`, but whether this is due to a) the method itself, b) fewer cell types in the `EPIC` signature gene lists, or c) marker gene information in the signature gene matrices themselves would require further investigation.
+
+
+Next, we can ask whether there is any relationship between these quantities:
```
allyhawkins, line 407: Given the differences in references, I'm not sure I would expect any relationship between the other quantities in the two methods. But I might expect to see a relationship between the shared cell types. You don't have to do this, but I do think it would be helpful to have a plot that directly compares the fractions of each of the shared cell types, either using a grouped sina plot or a scatter plot like the one below for each cell type.

allyhawkins (approved):
I think the new changes look good. I just added some of my own commentary that you might think about adding in to the conclusions and then one suggestion for an additional plot looking at shared cell types. It's up to you if you want to include it or not.

analysis/bulk-deconvolution/exploratory-notebooks/compare-epic-quantiseq.Rmd
sjspielman: >  But I might expect to see a relationship between the shared cell types.

sjspielman:
Noting that I did add the plots (basically no relationship between methods 😬 ) and some additional interpretation based on reviews.

PR #123, opened 2025-01-22
Recompile Figure 1

### LINKED ISSUE #90: Regenerate Figure 1 and supplementary tables with updated sample numbers

allyhawkins:
Closes #90 

This PR recompiles figure 1 with the updated panels for 1A and 1B (updated in #106). Nothing really fancy here, I just added in the new panels and adjusted the fonts to match what was there previously.


sjspielman (commented):
(edit for correct panels!)

The text in the "facet" header for panel ~C~ B looks kind of squished (it sort of looked that way before too it seems), but that's worth fixing! It's an aspect ratio problem.


I also actually wonder if we want to update panel ~B~ C since there is now an additional pub associated with this project (see attached screenshot), and both pubs are fully published. There are some other project card differences though we may not want to display though ("1 more sample will be made available soon") but that can be cropped out. Not 100% sure what specifics were done to create this screenshot panel in the first place, but do those again I guess!

Screenshot 2025-01-22 at 11 16 10 AM

allyhawkins:
1B text should be fixed and I updated the project card for 1C. It's a SVG that we can edit so I updated the publications and added the new metadata button.

sjspielman (approved):
Almost there, but don't need to see again so approving now! 

- ~Have a look the pubs, you've got some copy/paste problems - one is partially a duplicate of the other. You'll catch it!~ Nope, I am the problem! They are just back-to-back papers and you're correct! Ignore this bullet 😄 
- I'd suggest also adding a `...` at the end of "Additional sample metadata fields", similar to what we do for diagnosis and abstract, since there are more columns there in the real project card.

allyhawkins:
> I'd suggest also adding a ... at the end of "Additional sample metadata fields", similar to what we do for diagnosis and abstract, since there are more columns there in the real project card.

This isn't included since on a real version of the project card there is no `...`

Issue #128, opened 2025-01-28
Prepare normalized data for predicting bulk from pseudobulk


sjspielman:
We would like to ask if pseudobulk single-cell expression is predictive of bulk single-cell expression. To the extent that there is predictive power here, we want to further ask if any outliers (if detected) form meaningful biological groupings that provide insight into differences between bulk and single-cell modalities.  

The first step to this will be to prepare data for analysis:
- For bulk, we want TPM (potentially in log space). We've previously extracted these TSVs but with gene symbol ids, so this time we'll want to keep ensembl around
- For single-cell, we have a few choices for how to prepare the data. We'll want to calculate these versions and visualize them (shape of distribution, relationship to bulk) to determine which might serve us best. For the first two approaches below that don't account for gene lengths in mind, we should include that info in viz to assess whether the lack of consideration is causing bias.
	- We can sum raw counts to create pseudobulk and log-transform the counts with DEseq
	- We can sum logcounts directly 
		- Noting this tidbit from Aaron Lun [from a totally different context](https://github.com/SingleR-inc/SingleR/issues/3#issuecomment-519391898 )  about summing these values: _You may or may not prefer to sum the normalized expression values instead of the raw counts. The "advantage" of this approach is that the pseudo-bulk sample is not dominated by large cells, but the disadvantage is that you increase noise from small cells. (For DE analyses, there are more effects in play that discourage this practice, based on the total RNA content and mean-variance issues.)_ 
	- We can sum raw counts to create pseudobulk and calculate TPM from them, using gene lengths from the reference we use in `scpca-nf` (and then potentially in log space)

I anticipate two PRs for this issue:
- First, establish a new directory `analysis/prediction-from-pseudobulk` (other names super welcome!) with bulk TPM:
	- I plan to copy over some code from `analysis/bulk-deconvolution/` for this directly, but this time the script for TPM will keep ids in ensembl which we love
- Second, add a notebook to compare approaches for calculating the pseudobulk expression values. I plan to include all projects here, but if things look real weird we may have to go one project at a time.

sjspielman:
Noting that during a group discussion about this notebook, we decided to proceed with:
- pseudobulk calculated by summing raw counts and transforming with DESeq2
- bulk raw counts transformed with DESeq2

PR #133, opened 2025-01-28
Set up pseudobulk analysis and TPM data

sjspielman:
First half of #128 

This PR adds a new directory for the pseudobulk prediction with the following innards:

- R script to sync S3 files including both `quant.sf` and `_processed.rds`
  - For this analysis, it seems to me that we definitely can't use cellhash libraries unless we're prepared to actually perform demultiplexing which I do not think is on the agenda. Therefore, this parsing is different from the deconvolution version since we have to remove those samples. I also realized just now that a couple of the bulk samples were not actually paired with single-cell, so we lose a few more bulks there (only 6), but we end up with a reasonable enough N=104 samples overall:
    - `SCPCP000001`: 18 samples
    - `SCPCP000002`: 20 samples
    - `SCPCP000006`: 39 samples
    - `SCPCP000009`: 3 samples (the available non-multiplexed samples)
    - `SCPCP0000017`: 24 samples
- R script to calculate TPM for bulk and export as TSV
- Bash script to run the analysis

Locally, I have run the bash script and confirmed that quant.sf & rds files for all N=104 samples were synced, and TPM TSVs were calculated.


analysis/pseudobulk-bulk-prediction/scripts/sync-data-files.R
```diff
@@ -0,0 +1,137 @@
...
+# we do this first since there are fewer single-cell than corresponding bulk
+sync_sc_df <- library_metadata |>
+  dplyr::filter(
+    scpca_sample_id %in% solid_bulk_samples,
+    seq_unit %in% c("cell", "nucleus")
+  ) |>
+  dplyr::mutate(
+    s3_dir = file.path(s3_processed_path, scpca_project_id, scpca_sample_id),
+    output_dir = file.path(opts$output_dir, scpca_project_id, scpca_sample_id)
+  ) |>
```
jashapiro, line 85: because of how `aws sync` works, you don't really need either of these; you can just give it the base dir and let it find the correct library files based on the `include` pattern. But it isn't going to hurt either really, and it _is_ needed in the next one, so.

analysis/pseudobulk-bulk-prediction/scripts/sync-data-files.R
```diff
@@ -0,0 +1,137 @@
...
+stopifnot(
+  "Bulk and single-cell samples don't match" = setequal(sync_sc_df$scpca_sample_id, sync_bulk_df$scpca_sample_id)
+)
+
+
+
+# Sync quant.sf files ------------------
+sync_bulk_df |>
+  purrr::pwalk(
+    \(scpca_sample_id, output_dir, s3_dir) {
```
jashapiro, line 122: `scpca_sample_id` is unused.

jashapiro (approved):
LGTM

Issue #136, opened 2025-02-04
Explore bulk ~ pseudobulk model: Project 1


sjspielman:
Broken out of #129 

This issue tracks initiating the analysis notebook in the `bulk ~ pseudobulk` analysis for project `SCPCP000001`, following analysis steps delineated in [this comment](https://github.com/AlexsLemonade/scpca-paper-figures/issues/129#issuecomment-2634692446).

I plan to file PRs as I build up the notebook. This step-wise building up will also inform us whether one notebook per project is appropriate, or whether we need to divvy up even further based on specific analyses we are conducting. 

~1. Construct and select linear model~ (edited to cross off; see https://github.com/AlexsLemonade/scpca-paper-figures/issues/136#issuecomment-2637543879)
2. Second, use model residuals for GSEA with `hallmark`, `C7 ImmuneSigDB subcollection`, and `C8 cell type signature` MSigDB sets
3. Explore outliers (each is probably a PR as they may require threshold exploration): 
    - compare genes with residuals that are eg >=3`\sigma` across samples to assess whether we have shared genes
    - what are the genes with high expression in one modality but not the other? is there a relationship between these gene identities and wetlab considerations (prep kit, technology, sequencing unit)? where do they fall in the residual distribution?

It's important to note that we expect this project to have a decently predictive model, where other ScPCA projects (notably SCPCP000009 (cns) which has only 3 samples, and SCPCP000017 since working with bone is hard) may not have similar performance.

sjspielman:
Following some additional discussion https://github.com/AlexsLemonade/scpca-paper-figures/pull/138, we're going to back up and instead first model all projects in a single notebook, and then proceed with project-specific analysis notebooks next. 

Therefore, the scope of this issue should instead focus on steps 2 and 3 listed above, not step 1 which is taking place separately. This notebook will therefore begin by reading in the model to actually use for analysis.

PR #138, opened 2025-02-05
Initiate project 1 modeling

sjspielman:
Part 1 of #136 

This PR begins looking at project 1 by establishing a notebook with a model. I fit three models: without controlling for sample, one with `+sample_id` and one with `*sample_id`, and compare their fits a little bit to identify which to proceed with. They really are about the same all around so I'm opting to go with `+sample_id`. Let me know if this seems reasonable to you!
Edit: here's the HTML, but there isn't much there yet [SCPCP000001_bulk-pseudobulk.nb.html.zip](https://github.com/user-attachments/files/18673343/SCPCP000001_bulk-pseudobulk.nb.html.zip)

I did include a `utils.R` file to store functions used across notebooks which is sourced (with apologies) but documented! Currently, it just holds one function to read and prepare data, and it will get built up as we proceed here.

I plan to run a residual GSEA analysis for the next PR as described in #136.


analysis/pseudobulk-bulk-prediction/model-notebooks/SCPCP000001_bulk-pseudobulk.Rmd
```diff
@@ -0,0 +1,141 @@
...
+exclude_samples <- c("SCPCS000003")
+```
+
+
+### Paths
+
+```{r paths}
+analysis_dir <- here::here("analysis", "pseudobulk-bulk-prediction")
+data_dir <- file.path(analysis_dir, "data")
+metadata_dir <- file.path(rprojroot::find_root(rprojroot::is_renv_project), "s3_files")
```
jashapiro, line 38: Is this different from:
```suggestion
metadata_dir <- here::here("s3_files") 
```

analysis/pseudobulk-bulk-prediction/model-notebooks/utils.R
```diff
@@ -0,0 +1,47 @@
...
+      bulk = bulk_counts,
+      ensembl_id = gene_id
+    )
+
+  pseudo_df <- readr::read_tsv(pseudobulk_file, show_col_types = FALSE) |>
+    dplyr::filter(expression_type == "pseudobulk_deseq") |>
+    dplyr::select(ensembl_id, sample_id, pseudobulk = expression)
+
+  # combine bulk and pseudobulk into a single data frame
+  expr_df <- dplyr::full_join(
```
jashapiro, line 39: I don't think we need to include NA rows?
```suggestion
  expr_df <- dplyr::inner_join(
```

analysis/pseudobulk-bulk-prediction/model-notebooks/utils.R
```diff
@@ -0,0 +1,47 @@
...
+    purrr::pluck(project_id) |>
+    # make names consistent with pseudobulk df
+    dplyr::rename(
+      bulk = bulk_counts,
+      ensembl_id = gene_id
+    )
+
+  pseudo_df <- readr::read_tsv(pseudobulk_file, show_col_types = FALSE) |>
+    dplyr::filter(expression_type == "pseudobulk_deseq") |>
+    dplyr::select(ensembl_id, sample_id, pseudobulk = expression)
```
jashapiro, line 36: Filter sooner for efficiency (assuming `inner_join`)
```suggestion
    dplyr::select(ensembl_id, sample_id, pseudobulk = expression) |>
    dplyr::filter(!(sample_id %in% exclude_samples))
    
```

analysis/pseudobulk-bulk-prediction/model-notebooks/utils.R
```diff
@@ -0,0 +1,47 @@
...
+#' @param exclude_samples Optional vector of samples to remove from the final data frame
+#'
+#' @returns
+#' @export
+#'
+#' @examples
+prepare_data <- function(
+    bulk_counts_file,
+    pseudobulk_file,
+    project_id,
```
jashapiro, line 21: If you are going to map across projects (which I think you should) it may be easier with `project_id` first.
```suggestion
prepare_data <- function(
    project_id,
    bulk_counts_file,
    pseudobulk_file,
```

jashapiro (commented):
The model analysis here seems to make sense, but I think it makes sense to go ahead and do the overall stats for all projects before making a final decision. I don't expect the model choice to change (additive seems correct to me as a reasonable choice), but I think it makes sense to make the choice based on all projects and then not have all the model choice code there when you are getting into evaluation of each project. 

On the models to use though, I expected you do to use a random effects model for sample? I wouldn't expect the fit to change, but we definitely don't need to know the significance of each sample effect.

analysis/pseudobulk-bulk-prediction/model-notebooks/SCPCP000001_bulk-pseudobulk.Rmd
sjspielman: It is not, as (cc stephanie) this is not an openscpca module.

analysis/pseudobulk-bulk-prediction/model-notebooks/SCPCP000001_bulk-pseudobulk.Rmd
```diff
@@ -0,0 +1,141 @@
...
+purrr::map(models, summary)
+```
+
+In the end, we have:
+
+| model              | $R^2$     | $\sigma$ |
+|:-------------------|:----------|:---------|
+| `fit_plain`        | $0.7291$  | $2.332$  |
+| `fit_additive`     | $0.7292$  | $2.332$  |
+| `fit_interaction`  | $0.7293$  | $2.331$  |
```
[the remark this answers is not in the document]
jashapiro: 😆 I'm not saying you need to use them to make the final decision, but showing them in a table seems reasonable to me.

sjspielman:
@jashapiro Thanks for the feedback! My plan next is to update this notebook to have a section per project where I build/eval models for each. For this:
- I will read in metadata too and check for additional project-specific covariates we might want (there weren't any for project 1 that weren't also fully identifiable from sample)
- I'll bring in `lme4` as well for random effects - agreed on this point. 
- It's likely 1-2 more functions will make their way into `utils.R` too.

At the end of the notebook, I'll export an RDS with a list of all the models to proceed with.

jashapiro:
I was actually thinking you would not need a separate section for each project, but that you would be able to map across all projects and produce a table of fit statistics for the two models.

However, I had not considered this point:

> * I will read in metadata too and check for additional project-specific covariates we might want (there weren't any for project 1 that weren't also fully identifiable from sample)

I'm not sure what covariates you are specifically thinking about here, aside from single cell/single nuclei, which is probably always going to be worth including where present. 

What I was envisioning was something like (pseudocode, obviously):

```
map(projects){
  read(project)
  map(models){
    fit(project, model) |>
      glimpse()
  } |>
    rbind() # models
} |>
rbind() # projects 
```

Then you can print a single table with all the stats across projects, or produce some summary plots, etc. 

The idea here is to decide which model to use first, and then you will return to the project-by-project analysis, with only that model.

sjspielman:
> I'm not sure what covariates you are specifically thinking about here, aside from single cell/single nuclei, which is probably always going to be worth including where present.

It turns out this is not present, so that's all set!
Screenshot 2025-02-05 at 10 30 41 AM

Similarly, within a project, there is no variation in the 10x version. There is, however, a divide in the bulk where we see that project 1 has 2 single-end samples and the rest are paired-end. Library prep is consistent within project though:
Screenshot 2025-02-05 at 10 34 46 AM

The two single-end samples are `SCPCS000007` and `SCPCS000008`, which from the comparison notebook are unremarkable compared to the other samples; and also it's just two samples which makes considering its variation hard... So, I think we can consider technology/library items here safe to not consider in models. That takes care of that!

The main other variable I had been thinking we might want to include is `subdiagnosis` for project 6 specifically, since we have this split in samples:
Screenshot 2025-02-05 at 10 31 02 AM

sjspielman:
@jashapiro since the scope of this PR has changed, I'm going to close it and open a fresh one (incoming after I break for lunch).

sjspielman:
> The main other variable I had been thinking we might want to include is subdiagnosis for project 6 specifically, since we have this split in samples:

for posterity: nope, doesn't work since 1:1 with sample_ids even as a random effect. This means we're just controlling for sample all around.

PR #141, opened 2025-02-06
Modify model notebook parsing and export

### LINKED ISSUE #140: Also export project-specific data frames in modeling notebook
> I was starting next steps here just as you were approving, and I realize we'd benefit from some rearrangement in this notebook. We're going to actually need the gene ids that are in the models! The model objects don't save the full input data frame but only the columns that went into the model, so we should take this time to also export project-specific data frames with both expression flavors.
> 
> Therefore, I think we should split this into two steps:
> ```
> all_models <- purrr::map2(
>  pseudobulk_files, 
>  bulk_counts,
>  \(pseudo_file, bulk_df) {
>    prepare_data(pseudo_file, bulk_df, exclude_samples) |>  ### this should be a standalone map
>      model_data()  ### and this should be a standalone map
>  })
>```
>
> First, prepare the data, ~drop their NAs/Infs~, and then model the data. Then, we have the data frames with gene ids present in the environment, and we can include the project-specific data frame in the final exported RDS. I'll do this in a PR next.
>
_Originally posted by @sjspielman in https://github.com/AlexsLemonade/scpca-paper-figures/issues/139#issuecomment-2640144560_
            

Edit: I am reminded we are living in DESeq2 transformation land, so there are no NA/Infs anymore 😄

sjspielman:
Closes #140 

This PR updates the modeling notebook to prepare data and model in separate steps so that we can also export the _full_ underlying data including ensembl ids. I now export a list of `data` and `model`.
As I write this PR, I realize a different way to do this could be to take advantage of the `frame` slot in the model object which holds a data frame of the model variables. I could replace that data frame with full underlying data frame that includes the ensembl, so we would only need to export the model object in the end and not a list of two objects. Let me know if that would be preferable!

I also made two other small changes:
- Added a README to `results/`
- Removed a line snuck through in the notebook that was used for dev (the extra `readr::read_rds()` in there!).


sjspielman:
> Let me know your plans; I will approve with the small changes if you do not intend to put the ids into the model frame.

I do not intend to make this change for two reasons: i) keeping data separate doesn't make the objects that much bigger (the `lme4` object is already doing a good job making these big on its own!), and ii) I think it's probably slightly better for eventual data/results sharing to keep the model objects the same as `lmer4` returned them.

But I have pushed the requested changes

jashapiro (approved):
LGTM

Issue #144, opened 2025-02-07
Update data prep and models to filter infrequent genes


sjspielman:
This issue arose out of discussion in #142, specifically [this comment](https://github.com/AlexsLemonade/scpca-paper-figures/issues/142#issuecomment-2643722703):

> That makes me worry that something is going wrong earlier. I would never expect all samples in a project to have only zero or non-zero expression.
> 
> As far as >0 goes: the transformation is ~ log1p, but with size factors, so when you get < 0, it is probably a library scaling thing. The "right" thing to do would be to base this filter on the raw counts, though.

-----

I have explored some of the raw counts before transformation with DESeq, and I indeed detect genes that are expressed in some but not other samples! This is _much_ more in line with expectations, but it just is not apparent after the transformation. 

As such, we want to filter genes from on each project based on _raw counts_ before modeling. There are a couple filtering approaches we have considered:
- Remove genes which have zero expression in both bulk and pseudobulk
- Remove genes which, across both modalities, are expressed in fewer than X% of samples per project. We will probably start with 25% here. 

To this end, we will make the following changes, where each checkbox will be a PR:
- [ ] In the scripts that prepare input data that undergoes DESeq transformation (`scripts/calculate-pseudobulk.R` and `scripts/prepare-bulk-counts.R`), we'll also export a table with the fraction of samples that each gene is expressed in.
	- By using the actual sample counts and not logical we then leave open the possibility of each filtering approach listed above.
- [ ] In the notebook where we build models (`model-notebooks/build-assess-models.Rmd`), we'll read in these tables and use them for filtering before constructing models
	-  One idea here is also to use a notebook parameter to determine the filtering threshold, defaulting to 1 (i.e., remove genes that are expressed in fewer than 100% of samples)

PR #145, opened 2025-02-10
Export tables of gene expression percentage across samples

sjspielman:
Towards #144 

This PR updates data preparation scripts to also export tables with the percent of samples each gene is expressed in. These can be used for later filtering. Currently I am exporting these to `data`, but we could send them to `results` instead.

While I was here, I caught a few tiny bugs or mediocre code approaches (e.g. a hardcoded file path that should be an argument!) and made updates accordingly. In particular, I had added various files over time to `data/scpca_data` which somewhat broke the `run_prediction.sh` script for look over projects in that directory. I therefore decided to move files were are not literally ScPCA out of that directory and into its parent `data`. 

I also made a couple file names a bit more consistent with underscores/dashes. Because of these small filename changes, I also updated paths/file parsing in the exploratory notebook accordingly.


sjspielman:
@jashapiro - FYI I don't think pre-planning discussion really changed our plans for moving forward with this (and a second stacked PR that updates the modeling notebook with strategies for filtering). This way we can compare next steps at different filtering thresholds, which I do think it informative. For example, it has some potential to affect distributions of annotated vs non-annotated genes.

jashapiro (approved):
Looks fine, except that I don't like calling fractions percents.

PR #149, opened 2025-02-11
Update model notebook with filter parameter

### LINKED ISSUE #144: Update data prep and models to filter infrequent genes

sjspielman:
Closes #144 
This PR updates the modeling notebook to include a new parameter `filter_genes`. I ended up (at commit 1d9f8335355657d71a9e514c8e0fddbc15465a34) deciding to just keep this as boolean for now: If FALSE, we do not filter, and if TRUE, we filter to remove all genes with zero expression in both modalities. In other words, I don't currently have code here to remove genes based on a threshold fraction of samples it's expressed in, since thresholds are hard! Also, when I ran with a threshold, the interaction model just plain errorred and the code got a bit meandering to work around that. I figure if we really want to use other filtering options, we might have to back to the cell-level anyways...so for now let's just kick that can down the line!

In addition to updating the notebook to perform filtering where necessary, I also updated text in the notebook and the oob squish limit to ensure it is compatible with both parameter options. The notebook is now rendered twice per parameter with a separate HTML exported each time. Also some small changes for file names I fiddled with in #145. `renv` also got updated since params bring shiny.


analysis/pseudobulk-bulk-prediction/model-notebooks/.gitignore
```diff
@@ -0,0 +1 @@
+build-assess-models.nb.html
```
sjspielman, line 1: sad.
```suggestion
build-assess-models.nb.html

```

analysis/pseudobulk-bulk-prediction/model-notebooks/build-assess-models.Rmd
```diff
@@ -60,44 +70,82 @@ Define files to read in:
...
-prepare_data <- function(pseudo_file, bulk_df, exclude_samples) {
+prepare_data <- function(pseudo_file, bulk_df, id, exclude_samples, expressed_df, filter = params$filter_genes) {
+  
+  # Find genes to keep if filtering is on
+  if (filter) {
+    keep_genes <- expressed_df |>
+      dplyr::filter(
+        project_id == id, 
+        pseudobulk_frac_expressed > 0,
+        bulk_frac_expressed > 0
```
jashapiro, line 119: Your current filter requires both bulk and pseudobulk to be expressed, which is not what we would want, as we could be quite interested in genes that are expressed in one but not the other. 
```suggestion
        pseudobulk_frac_expressed > 0 | bulk_frac_expressed > 0
```

That said, I don't think I agree with your decision to implement this as a boolean. If you instead made the filter numeric, you could do something as simple as:


```
pseudobulk_frac_expressed + bulk_frac_expressed > expr_threshold * 2
```

where setting having `expr_threshold = -1` would be equivalent to no filtering, and `expr_threshold = 0` would be the same as the above filter of expression in at least one sample. If we want to return with more filters, then we can do that later with no additional code changed in this script.

(Note that this also removes the need for the entire `if` block.)

sjspielman: Sure, we can do this! The only thing that might get us is the interaction models may not run with a different threshold, but I say we cross that bridge if/when we ever get there?

jashapiro: I don't understand why it should have affected things at all. The filtering based on genes is across all samples, so the matrices are still complete.

sjspielman: Frankly I didn't understand it myself but unfortunately don't immediately have the exact code under which it occurred. Maybe it was a different bug altogether then...

sjspielman:
This has now been updated to use the threshold parameter instead of boolean.

analysis/pseudobulk-bulk-prediction/model-notebooks/build-assess-models.Rmd
```diff
@@ -17,6 +19,12 @@ We build and assess the following random effects models for each project:
 * `fit_additive`: `bulk ~ pseudobulk + (1|sample_id)`
 * `fit_interaction`: `bulk ~ pseudobulk + (1|sample_id) + (1|sample_id:pseudobulk)`
 
+
+In addition, the notebook parameter `expr_threshold` determines how genes are filtered from the data before modeling: Only genes which are expressed in at least `2 * expr_threshold` in both modalities combined are retained.
```
jashapiro, line 23: The wording here is tough, but I think it is worth trying to describe the intent rather than the precise calculation.

But we definitely want to say "greater than" rather than "at least" because filtering at 0 does filter samples!
```suggestion
In addition, the notebook parameter `expr_threshold` determines how genes are filtered from the data before modeling: Only genes which are expressed in greater than the proportion `expr_threshold` of samples, averaged across bulk and pseudobulk modalities, are retained.
```

analysis/pseudobulk-bulk-prediction/model-notebooks/build-assess-models.Rmd
```diff
@@ -60,44 +68,76 @@ Define files to read in:
...
 This chunk defines a helper function to read and combine bulk and pseudobulk datasets for a single project.
 ```{r}
-prepare_data <- function(pseudo_file, bulk_df, exclude_samples) {
+prepare_data <- function(pseudo_file, bulk_df, id, exclude_samples, expressed_df) {
+  
+  # determine genes to keep
+  keep_genes <- expressed_df |>
+    dplyr::filter(
+      project_id == id, 
+      (pseudobulk_frac_expressed + bulk_frac_expressed) > (params$expr_threshold * 2)
```
jashapiro, line 115: Since I was saying something about the intent, we might as well put that in the code more directly.
```suggestion
      mean(pseudobulk_frac_expressed, bulk_frac_expressed) > params$expr_threshold
```

jashapiro, line 115: Since I was saying something about the intent, we might as well put that in the code more directly.
```suggestion
      mean(pseudobulk_frac_expressed, bulk_frac_expressed) > params$expr_threshold
```

analysis/pseudobulk-bulk-prediction/model-notebooks/build-assess-models.Rmd
```diff
@@ -60,44 +68,76 @@ Define files to read in:
...
+  file.path(data_dir, "fraction-expressed-bulk.tsv")
+) |>
+  dplyr::rename(bulk_frac_expressed = frac_samples_expressed)
+  
+named_projects <- purrr::set_names(pseudobulk_names, pseudobulk_names)
+expressed_df <- named_projects |>
+  purrr::map(
+    \(project_id) {
+      readr::read_tsv(
+        file.path(data_dir, glue::glue("{project_id}_fraction-expressed-single-cell.tsv"))
```
jashapiro, line 95: Avoid repeated col_type printing.
```suggestion
        file.path(data_dir, glue::glue("{project_id}_fraction-expressed-single-cell.tsv")),
        show_col_types = FALSE
```

analysis/pseudobulk-bulk-prediction/model-notebooks/build-assess-models.Rmd
```diff
@@ -60,44 +68,76 @@ Define files to read in:
...
 bulk_counts <- readr::read_rds(bulk_counts_file)
 
 # Make sure we have the same projects, in the same order
 stopifnot(
   all.equal(names(pseudobulk_files), names(bulk_counts))
 )
+
+# also the tables with fraction expressed
+bulk_frac_expr_df <- readr::read_tsv(
+  file.path(data_dir, "fraction-expressed-bulk.tsv")
```
jashapiro, line 86: may as well skip this one too.
```suggestion
  file.path(data_dir, "fraction-expressed-bulk.tsv"),
  show_col_types = FALSE
```

analysis/pseudobulk-bulk-prediction/run-prediction.sh
```diff
@@ -57,5 +57,17 @@ for project_dir in $scpca_dir/*; do
       --output_frac_expressed_file "${fraction_expressed_file}"
 done
 
-# Build and export models to results/models/
-Rscript -e "rmarkdown::render('${model-notebooks}/build-assess-models.Rmd')"
+# Build and export models to results/models across different thresholds for expression
+for expr_threshold in -1 0; do
```
jashapiro, line 61: Just out of curiosity, did you try a small threshold? I'm mostly curious if it works unmodified. If it does not, I would put a note in the notebook where you define the parameter to note that values > 0 don't work (and/or add code to remove the interaction model for those cases).
```suggestion
for expr_threshold in -1 0 0.1; do
```

jashapiro (approved):
This looks fine, though I am curious if the model really fails with a small positive filter.

My main other comments are about wording, and removing `read_tsv` chattiness.

analysis/pseudobulk-bulk-prediction/model-notebooks/build-assess-models.Rmd
sjspielman: thank you! boy was I having a fun time phrasing this, which is why I gave up and just showed the math

analysis/pseudobulk-bulk-prediction/model-notebooks/build-assess-models.Rmd
jashapiro: Note updated code above...

analysis/pseudobulk-bulk-prediction/run-prediction.sh
```diff
@@ -57,5 +57,17 @@ for project_dir in $scpca_dir/*; do
...
+    threshold_str="all-genes"
+  else
+    threshold_str="threshold-${expr_threshold}"
+  fi
+
+  Rscript -e "rmarkdown::render('${model_notebook_dir}/build-assess-models.Rmd',
+              params = list(expr_threshold = ${expr_threshold}),
+              output_file = 'build-assess-models_${threshold_str}.nb.html',
+              output_dir = '${model_notebook_dir}')"
+done
```
sjspielman, line 73: oh no i got got! i'll appease github in the next PR.

PR #150, opened 2025-02-11
Explore residual distributions across annotation status

### LINKED ISSUE #146: Compare residuals between annotated and unannotated genes
One question that has come up during GSEA analysis of residuals is whether we should be filtering genes based on their "realness" as it were - In other words, we expect a lot of the genes in a given genome reference are small, infrequently expressed transcripts of not-totally-"real" gene status, and their inclusion may be confounding results.

One way to test this is to use gene annotation as a proxy: We can compare distributions of genes that have annotated gene symbols (bona fide real genes) to those which don't (expressed but maybe not as "real" of genes). If we see major differences, this is an indicator that the presence of these changes is biasing results.

Additional analysis details: 
- We may wish to do this analysis at different levels of pre-filtering genes to assess if this potential confounder is independent of any filtering we perform based solely on low-zero gene expression or not.
- We might also look at the extent to which genes in the MSigDB gene sets of interest are annotated or not. Hopefully they are indeed annotated!

sjspielman:
Closes #146 
Stacked on #149 

This PR does what it says! Findings:
- yes, residuals distribution are higher in annotated genes. This means the variation of interest is more likely to be in the annotated genes.
- The difference actually increases for the residuals built from the filtered version of the data. It seems after filtering using only the approach pending in #149, we reduce our overall share of annotated (biologically more likely to be meaningful) genes.

Here's the rendered notebook:
[residuals-annotated-genes.nb.html.zip](https://github.com/user-attachments/files/18759712/residuals-annotated-genes.nb.html.zip)


analysis/pseudobulk-bulk-prediction/exploratory-notebooks/residuals-annotated-genes.Rmd
```diff
@@ -0,0 +1,190 @@
...
+    }) |>
+    purrr::list_rbind(names_to = "project_id")
+}
+```
+
+This chunk defines a helper function for plotting boxplots.
+```{r}
+make_boxplot <- function(df, yvar, title) {
+  ggplot(df) +
+    aes(x = sample_id, fill = status, y = {{yvar}}) +
```
jashapiro, line 75: I don't think I care too much about the per-sample distributions. I would rather see the overall distribution per project in more detail than a boxplot (e.g., histogram).

jashapiro (commented):
The overall form here looks good, but my main comment is that I don't think the per-sample boxplots are the best way to look at this. I'd rather see some kind of histogram or density plot to visualize each project.

sjspielman:
@jashapiro thanks for the feedback here! I actually started these with project-wide density plots which were really hard to make out before I plotted at the sample level, so this time I'll try project-wide ridge plots or histograms and pick the one that is most readable.

analysis/pseudobulk-bulk-prediction/exploratory-notebooks/residuals-annotated-genes.Rmd
```diff
@@ -0,0 +1,190 @@
...
+```
+
+
+One wonders, what's the ratio of annotated to not annotated genes for each version of the data (unfiltered and filtered)?
+The unfiltered should be the same for all projects, but filtered will be different -
+
+```{r}
+unfiltered_df |>
+  dplyr::group_by(project_id) |>
+  dplyr::count(status) |>
```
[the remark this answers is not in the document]
jashapiro: updated the above suggestion to remove separate `group_by`, which isn't needed.

sjspielman:
Next round has landed, with overlaid project histograms instead of per-sample boxplots. I also tightened some code up to make the notebook more modular with functions to wrap the modeling and the ratio calculations.

> Also, worth checking that the numbers here should be the same? At least for the pseudobulk...

I don't think we can _directly_ do this check, because the data incoming into this notebook only considers the genes that actually went into the models. This means data was already subset to the shared genes between bulk and pseudobulk, so `ensembl_symbol_df` will come up as having more genes than are in our data. Therefore, to enable a more robust check here, I grabbed a TPM file to first filter the `ensembl_symbol_df` to only bulk genes, which is expected match the genes in our data - and indeed! This also confirms that if we were to removed unannotated genes before modeling, we'd be left with _at least_ 40287 genes, and even less if we apply expression filters. That's still rather a lot of genes!

Here's the updated notebook:
[residuals-annotated-genes.nb.html.zip](https://github.com/user-attachments/files/18770939/residuals-annotated-genes.nb.html.zip)

analysis/pseudobulk-bulk-prediction/exploratory-notebooks/residuals-annotated-genes.Rmd
```diff
@@ -0,0 +1,238 @@
...
+    purrr::list_rbind(names_to = "project_id")
+}
+```
+
+This chunk defines a function for plotting histograms.
+```{r}
+make_histogram <- function(df, xvar, title) {
+  ggplot(df) +
+    aes(x = {{xvar}}, fill = status) + 
+    geom_histogram(alpha = 0.9, bins = 150) +
```
jashapiro, line 75: Did you try looking at density here? I think it might make the comparison easier. I might do somewhat fewer bins too?
```suggestion
    aes(x = {{xvar}}, y = after_stat(density), fill = status) + 
    geom_histogram(alpha = 0.5, bins = 100) +
```

Or with different scaling:
```suggestion
    aes(x = {{xvar}}, y = after_stat(ndensity), fill = status) + 
    geom_histogram(alpha = 0.5, bins = 100) +
```

sjspielman:
Using density stat helped, but not a ton - the spikiness is there as long as I plot the unfiltered data probably, but we also don't have to plot _this_ version of the data. 
[residuals-annotated-genes.nb.html.zip](https://github.com/user-attachments/files/18771982/residuals-annotated-genes.nb.html.zip)

analysis/pseudobulk-bulk-prediction/exploratory-notebooks/residuals-annotated-genes.Rmd
```diff
@@ -0,0 +1,237 @@
...
+
+This chunk defines a function for plotting histograms.
+```{r}
+make_histogram <- function(df, xvar, title) {
+  ggplot(df) +
+    aes(x = {{xvar}}, y = after_stat(ndensity), fill = status) + 
+    geom_histogram(alpha = 0.5, bins = 100) +
+    facet_wrap(vars(project_id), scales = "free", ncol = 1) +
+    scale_fill_brewer(palette = "Set1") +
+    ggtitle(title)
```
jashapiro, line 78: Something strange is going on here, as `ndensity` should scale to max of 1, but the `annotated` values are being scaled to 2.... oh, they are stacked.
```suggestion
  ggplot(df) +
    aes(x = {{xvar}}, y = after_stat(ndensity), fill = status) + 
    geom_histogram(position = "identity", alpha = 0.5, bins = 100) +
    facet_wrap(vars(project_id), scales = "free", ncol = 1) +
    scale_fill_brewer(palette = "Set1") +
    ggtitle(title)  
```

sjspielman: 🎉 we're in business, re-rendering notebook now. 
Screenshot 2025-02-12 at 12 57 35 PM

analysis/pseudobulk-bulk-prediction/exploratory-notebooks/residuals-annotated-genes.Rmd
```diff
@@ -0,0 +1,237 @@
...
+}
+```
+
+This chunk defines a function for plotting histograms.
+```{r}
+make_histogram <- function(df, xvar, title) {
+  ggplot(df) +
+    aes(x = {{xvar}}, y = after_stat(ndensity), fill = status) + 
+    geom_histogram(position = "identity", alpha = 0.45, bins = 100) +
+    facet_wrap(vars(project_id), scales = "free", ncol = 1) +
```
jashapiro, line 76: I find the shifts in the axes to be distracting, and the y axis is now fixed, so I think we can remove the free scales. 
```suggestion
    facet_grid(rows = vars(project_id)) +
```

If you want to set scales to avoid extra whitespace, you could add `coord_cartesian` calls when you use this.

analysis/pseudobulk-bulk-prediction/exploratory-notebooks/residuals-annotated-genes.Rmd
```diff
@@ -0,0 +1,237 @@
...
+
+```{r}
+unfiltered_df <- read_prepare_data("all-genes\\.rds$", ensembl_symbol_df) 
+```
+
+
+Do the residuals have different distributions based on gene annotation status?
+
+```{r, fig.height = 12, fig.width = 5}
+make_histogram(unfiltered_df, residuals, "residual distributions")
```
jashapiro, line 124: A thought for zooming a bit without losing any real info. 
```suggestion
make_histogram(unfiltered_df, residuals, "residual distributions") +
  coord_cartesian(xlim= c(-15, 15))
```

jashapiro (approved):
LGTM, though I still prefer aligned plots. I don't need to see this again.

PR #151, opened 2025-02-12
Notebook to run GSEA with shuffled gene sets

### LINKED ISSUE #147: Run GSEA with shuffled gene sets
In some initial attempts at running GSEA on the residual distributions, we have observed that _many_ gene sets are coming back significant; for example for some projects, >=40/50 of MSigDB hallmark pathways generally coming back significant. This seems like a lot! So, we'd like to take a step back and test if we get similar results shuffling genes. If we are getting similar numbers of significant pathways, this tells us that this analysis may not be worth more investment.

For this, we can run GSEA on residuals with a shuffled version of MSigDB's gene sets of interest such that gene ids are randomly shuffled among pathways, retaining the same total number of gene ids per pathway (edit) and compare to results using unshuffled, "correct" gene sets.

sjspielman:
Closes #147 
Stacked on #149 

This PR adds a notebook to run GSEA on the ranked median residuals, `exploratory-notebooks/permute-gsea.Rmd`. and I run this under a variety of conditions, controlled by Rmarkdown parameters and executed via the bash script `run_permute-gsea.sh`. Rendered notebooks across the following conditions are saved in`exploratory-notebooks/permute-gsea-htmls`:

* Both the `H` and `C8` MSigDB signature sets
* Three different versions of the residuals based on three different models constructed in #149. 
  * `all-genes`: All genes are used and no filtering is performed.
  * `threshold-0` and `threshold-0.25`: Genes are filtered such that, on average between modalities, only genes present in a`> threshold` fraction of samples are retained.
    * 0 corresponds to: gene is present in at least 1 sample
    * 0.25 corresponds: gene is present in at least >25% of samples

I ended up using 50 replicates as well for the shuffling; my original plan of 100 was taking a bit too long for what seemed like 
not much more information!

At the end of each report, I display histograms of the number of significant pathways returned per project in either enrichment direction (bulk or pseudobulk). I observed similar results to what we have seen with real data - for hallmark, 45-50 pathways are coming back enriched in bulk, and for C8, between 300-600 pathways are coming back enriched in bulk. However, I never observe enrichment towards pseudobulk. 

To help with review, here are those notebooks:
[permute-gsea-html.zip](https://github.com/user-attachments/files/18773613/permute-gsea-html.zip)

Something to think about for review: It occurred to me while filing this that it might be helpful context to actually go ahead and run the "correct" GSEAs in this notebook too, to have real and shuffled results side-by-side. Let me know if I should add that in!


jaclyn-taroni:
> Something to think about for review: It occurred to me while filing this that it might be helpful context to actually go ahead and run the "correct" GSEAs in this notebook too, to have real and shuffled results side-by-side. Let me know if I should add that in!

That does seem like it would be nice, but it really depends on how much work that would be. It seems like _GSEA is really not a good idea_!

analysis/pseudobulk-bulk-prediction/exploratory-notebooks/permute-gsea.Rmd
```diff
@@ -0,0 +1,193 @@
...
+Note that this data considers all results GSEA deemed to be signficant; there is no further filtering by effect size or P-value at this time.
+
+```{r, fig.width = 10, fig.height = 3, message = FALSE}
+ggplot(shuffled_gsea_df) + 
+  aes(x = count, fill = direction) + 
+  geom_histogram() + 
+  facet_wrap(vars(project_id), nrow = 1) +
+  scale_fill_brewer(palette = "Dark2") +
+  theme(legend.position = "bottom")
+```
```
jaclyn-taroni, line 186: Returning this while we ponder adding in real results. Can you give this a nicer x-axis label? Both axes are labeled with "count," and I found it confusing.

sjspielman:
> That does seem like it would be nice, but it really depends on how much work that would be. It seems like GSEA is really not a good idea!

I wouldn't take much to add in running GSEA and a barplot of the number significant gene sets per project, so I think it's worth adding in. Also, since we haven't actually sent code to main yet running GSEA in the first place, this seems like we get a two-for-one putting it all in this notebook (perhaps renaming the notebook so it doesn't seem like it _only_ permutes!).

Another FYI towards review - while looking back over my code here, I realized I'm almost certainly not doing sampling correctly. I need to be preserving the number of times each gene appears in the shuffled version. I plan to fix this as well before re-requesting review; I don't expect results to meaningfully change but it would be nice to be surprised..!

jaclyn-taroni:
> I need to be preserving the number of times each gene appears in the shuffled version. 

I don't think I understand. Why?

sjspielman:
> I don't think I understand. Why?

Let's say there are X total genes across the signatures, Y of which are unique. (In other words, if I pull out the gene signature data frame from msigdbr, it's X rows long).
Currently what I am doing is: For each for each gene set, I am counting the number of genes N and sampling from the unique genes in Y. The reason I did this at first was because it helps me guarantee that we have unique genes in each gene set, which we are _not_ guaranteed if we naively just shuffle genes among gene set labels. But, my approach has _not_ guaranteed that the shuffled genes represent the same spread of X genes pooled across gene sets, which seems like a problem? (I hope some of these words made sense!)

jaclyn-taroni:
If the goal of the analysis is to ask: Do random gene sets of the same size also produce significant results? 

Then, it's not apparent to me why conserving how frequently an individual gene is in a gene set (provided I followed you correctly) matters?

sjspielman:
> If the goal of the analysis is to ask: Do random gene sets of the same size also produce significant results?

Ah well yes, the current code does achieve this goal! Maybe I'm overthinking... I wouldn't use this kind of approach for a formal permutation test probably, but we're not doing that here.

jaclyn-taroni:
To be fair, you wouldn't use 50 replicates for that either.

sjspielman:
I've updated the code here to run GSEA on real gene sets, and I updated the plots section with additional explanation to hopefully make the different plots easier to interpret. 
In addition, as this notebook now actually runs a GSEA, I would consider it part of analysis and not only exploratory. I therefore re-arranged a bit:
- I removed the script I initially added that runs this notebook, since now the notebook is run from `run-prediction.sh`.
- I moved this notebook out of `exploratory-notebooks` and into `model-notebooks`, which I also renamed to `notebooks` to reflect more generally that it contains notebooks that are part of analysis (not specifically building models)
- To reduce clutter in this directory, I added a subdirectory `model-htmls` where I moved the model notebook HTMLs into
- READMEs were updated accordingly

Since running this whole thing through takes a little bit, I thought I might send this back for code review first and then I can re-run all the notebooks. To support review, I have rendered a notebook with 10 replicates using the hallmark gene set so you can see how it looks. Please note in particular how I've updated the histogram.. have I gone too far and made it more confusing? 😬 
[perform-gsea.nb.html.zip](https://github.com/user-attachments/files/18786967/perform-gsea.nb.html.zip)

If you prefer to wait to see all the notebooks, no problem, let me know!

sjspielman:
@jaclyn-taroni here's an updated notebook with an interpretable plot 😄 (still n=5 reps)
[perform-gsea.nb.html.zip](https://github.com/user-attachments/files/18790559/perform-gsea.nb.html.zip)

analysis/pseudobulk-bulk-prediction/notebooks/perform-gsea.Rmd
```diff
@@ -0,0 +1,279 @@
...
+    # calculate fraction version
+    frac = count/num_genesets
+  )
+```
+
+## Compare GSEA to shuffled GSEA results
+
+Finally, we'll make plots comparing these two analyses:
+
+* A bar plot of the number of significant pathways per project detected with GSEA
```
jaclyn-taroni, line 224: This is my interpretation of the plot, anyway:
```suggestion
* A bar plot of the proportion of input pathways returned as significant per project detected with GSEA
```

analysis/pseudobulk-bulk-prediction/notebooks/perform-gsea.Rmd
```diff
@@ -0,0 +1,279 @@
...
+* A histogram of the number of significant pathways per project detected across the `r params$reps` replicates of shuffled GSEAs
```
jaclyn-taroni, line 225:
```suggestion
* A histogram of the proportion of input random gene sets returned as significant per project detected across the `r params$reps` replicates of GSEA
```

analysis/pseudobulk-bulk-prediction/notebooks/perform-gsea.Rmd
```diff
@@ -0,0 +1,279 @@
...
+  show_col_types = FALSE
+)
+```
+
+### Functions
+
+
+This chunk defines a function to create a data frame with a shuffled version of genesets.
+```{r}
+generate_shuffle <- function(geneset_count_df, ensembl_ids){
```
jaclyn-taroni, line 52: I think my preferred approach would be to have this function take the `geneset_df` and number of replicates and return _all of the random genesets_ instead of getting passed to `purrr::map()`. That strikes me as more straightforward to follow (and more self-contained, which are related concepts IMO) if I had to return to this after some months.

jaclyn-taroni (commented):
This looks mostly good. I have some nits to pick with the language used to describe the plots & interpretation, and if I think about returning to this analysis at some later date, I think restructuring `generate_shuffle()` is going to help with readability.

jaclyn-taroni:
Just to add: I don't think any of my comments necessarily prevent you from starting the Big Run™️ unless you want me to take another look at `generate_shuffle()`.

sjspielman:
I have updated the shuffle function to take a geneset and number of replicates, and it now returns a list of shuffled gene set data frames. 
Here are notebooks from the Big Run:
[gsea-htmls.zip](https://github.com/user-attachments/files/18803365/gsea-htmls.zip)

jaclyn-taroni (approved):
I'm going to approve this. It is interesting that sometimes there are significant pseudobulk results, and there appear to be no significant pseudobulk results for the shuffled gene sets. So, that suggests that there might be some positive correlation structure in the positive residuals that makes GSEA return significant results even for biologically meaningless sets of genes and differs from any kind of structure in the genes with negative residuals.

PR #156, opened 2025-02-14
Establish ESTIMATE analysis directory

### LINKED ISSUE #152: Establish analysis directory for ESTIMATE analysis
Following #148, we will plan to proceed with more formal ESTIMATE analysis. The issue tracks adding an analysis directory to run ESTIMATE across modalities. This PR will also contain a script to obtain relevant data files from S3 for analysis:
- quant.sf files (for TPM calculations)
- bulk raw count files 
- SCE files for pseudobulk calculation

We'll also export two files to help with mapping various items:
  - Bulk counts are labeled by the library ID, so we'll need to convert these to sample IDs
  - From the SCE, we'll need to grab a map of ensembl id <-> gene symbol, since ESTIMATE wants symbols (or entrez ids, but we have symbols!)

Much of this code will be borrowed from:
- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/pseudobulk-bulk-prediction/scripts/sync-data-files.R 
- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/bulk-deconvolution/scripts/prepare-data-files.R

sjspielman:
Closes #152 
This PR establishes the `analysis/estimate` directory with:
- a bash run script
- various readmes where appropriate
- a script to sync data files and export TSV maps we'll need for next steps. This script is largely the same as [this other sync script](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/pseudobulk-bulk-prediction/scripts/sync-data-files.R), except it also has parts [like this script](https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/bulk-deconvolution/scripts/prepare-data-files.R) to create the gene id map tsv.

All output so far is sent to `data/` whose contents are ignored, but it is present with `.gitkeep`. I also fixed a couple typos in the original script I modeled this one after that I found while I was here.

I have run the code locally and it ran as expected to sync files and create the map TSVs.

### DIFF
--- a/analysis/estimate/run-estimate-analysis.sh
+++ b/analysis/estimate/run-estimate-analysis.sh
@@ -19,11 +19,9 @@ mkdir -p $scpca_dir
 mkdir -p $tpm_dir
 mkdir -p $pseudobulk_dir
 
-library_sample_map_file="${data_dir}/bulk-library-sample-ids.tsv"
 ensembl_symbol_map_file="${data_dir}/ensembl-symbol-map.tsv"
 
 # Sync data files from S3
 Rscript ${script_dir}/sync-data-files.R \
   --output_dir "${scpca_dir}" \
-  --library_sample_map_file "${library_sample_map_file}" \
   --ensembl_symbol_map_file "${ensembl_symbol_map_file}"
--- a/analysis/estimate/run-estimate-analysis.sh
+++ b/analysis/estimate/run-estimate-analysis.sh
@@ -1,0 +1,29 @@
+#!/bin/bash
+
+# This script runs an analysis using ESTIMATE
+
+set -euo pipefail
+
+# Run script from its location
+basedir=$(dirname "${BASH_SOURCE[0]}")
+cd "$basedir"
+
+# Define directories
+data_dir="data"
+script_dir="scripts"
+scpca_dir="${data_dir}/scpca"
+tpm_dir="${data_dir}/tpm"
+pseudobulk_dir="${data_dir}/pseudobulk"
+
+mkdir -p $scpca_dir
+mkdir -p $tpm_dir
+mkdir -p $pseudobulk_dir
+
+library_sample_map_file="${data_dir}/bulk-library-sample-ids.tsv"
+ensembl_symbol_map_file="${data_dir}/ensembl-symbol-map.tsv"
+
+# Sync data files from S3
+Rscript ${script_dir}/sync-data-files.R \
+  --output_dir "${scpca_dir}" \
+  --library_sample_map_file "${library_sample_map_file}" \
+  --ensembl_symbol_map_file "${ensembl_symbol_map_file}"
--- a/analysis/estimate/scripts/README.md
+++ b/analysis/estimate/scripts/README.md
@@ -1,0 +1,8 @@
+This directory contains scripts used in the analysis to predict bulk expression from matched pseudobulk expression.
+
+* `sync-data-files.R`: This script syncs data files for analysis.
+  * It identifies solid tumor samples of interest libraries of interest and obtains their bulk `quant.sf` files and single-cell `_processed.rds` files from S3.
+  Files are saved in `../data/scpca/<project id>/<sample id>/`.
+  * It also exports two TSVs to `../data`:
+    * Map for bulk library and sample ids
+    * Map for ensembl ids and gene symbols
--- a/analysis/estimate/scripts/README.md
+++ b/analysis/estimate/scripts/README.md
@@ -1,8 +1,6 @@
 This directory contains scripts used in the analysis to predict bulk expression from matched pseudobulk expression.
 
 * `sync-data-files.R`: This script syncs data files for analysis.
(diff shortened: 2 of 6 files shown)


analysis/estimate/run-estimate-analysis.sh
```diff
@@ -0,0 +1,29 @@
...
+# Run script from its location
+basedir=$(dirname "${BASH_SOURCE[0]}")
+cd "$basedir"
+
+# Define directories
+data_dir="data"
+script_dir="scripts"
+scpca_dir="${data_dir}/scpca"
+tpm_dir="${data_dir}/tpm"
+pseudobulk_dir="${data_dir}/pseudobulk"
```
allyhawkins, line 16: I don't think these are used yet here?

analysis/estimate/scripts/sync-data-files.R
```diff
@@ -0,0 +1,203 @@
...
+    "--scpca_sce_file",
+    type = "character",
+    default = here::here("s3_files", "SCPCS000001", "SCPCL000001_processed.rds"),
+    help = "Path to an SCE file used to obtain mappings for converting ensembl ids to gene symbols"
+  ),
+  make_option(
+    "--library_sample_map_file",
+    type = "character",
+    help = "Path to output TSV file mapping bulk sample and library ids",
+  ),
```
allyhawkins, line 32: I know you do this in the other script, but I think you could skip this step here. We already have a script that you can run that will save the metadata for ScPCA to the `s3_files` folder (see https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/scripts/figure_setup/sync-metadata.R). I'm not sure why you can't use the library metadata and just filter to only bulk IDs in later steps.

analysis/estimate/scripts/sync-data-files.R
```diff
@@ -0,0 +1,203 @@
...
+    type = "character",
+    help = "Path to output TSV file mapping ensembl ids and gene symbols",
+  ),
+  make_option(
+    "--sample_metadata_file",
+    type = "character",
+    default = here::here("s3_files", "scpca-sample-metadata.tsv"),
+    help = "Path to ScPCA sample metadata file."
+  ),
+  make_option(
```
allyhawkins, line 44: Wait... you are using the library metadata file, so why are you making a separate map?

allyhawkins (commented):
This mostly looks good, but I had a question about why you are making a new library metadata file?

analysis/estimate/scripts/sync-data-files.R
sjspielman: This is a pretty good point.....! Will remove this map since we can just use the library metadata when we need it rather than making extra files.............

analysis/estimate/run-estimate-analysis.sh
sjspielman: No, they aren't use yet, but will be in next steps. I could delete these lines now but they'll just get added back in the next PR, so I think it's fine to keep here?

sjspielman:
Extra mapping file we don't need removed!

analysis/estimate/run-estimate-analysis.sh
allyhawkins: I'm good with keeping them, just wanted to check!

analysis/estimate/scripts/README.md
```diff
@@ -0,0 +1,8 @@
+This directory contains scripts used in the analysis to predict bulk expression from matched pseudobulk expression.
+
+* `sync-data-files.R`: This script syncs data files for analysis.
+  * It identifies solid tumor samples of interest libraries of interest and obtains their bulk `quant.sf` files and single-cell `_processed.rds` files from S3.
```
allyhawkins, line 4:
```suggestion
  * It identifies solid tumor libraries of interest and obtains their bulk `quant.sf` files and single-cell `_processed.rds` files from S3.
```
I think this is what you mean?

analysis/estimate/scripts/README.md
```diff
@@ -0,0 +1,8 @@
...
+  Files are saved in `../data/scpca/<project id>/<sample id>/`.
+  * It also exports two TSVs to `../data`:
+    * Map for bulk library and sample ids
```
allyhawkins, line 7:
```suggestion
```

allyhawkins (approved):
LGTM, just two docs updates that need to be fixed.

analysis/estimate/scripts/README.md
sjspielman: eh this one is tomato tomAto in terms of sample/library but indeed, only one of those words!

analysis/estimate/scripts/README.md
```diff
@@ -0,0 +1,8 @@
...
+    * Map for ensembl ids and gene symbols
```
sjspielman, line 8:
```suggestion
  * It also exports a TSV to `../data` that maps ensembl ids to gene symbols
```

PR #157, opened 2025-02-18
Add script to prepare data for estimate

### LINKED ISSUE #153: Add scripts to prepare input data for ESTIMATE
Following #152, the next step will be to add scripts to calculate input quantities for running ESTIMATE, including:
- bulk tpm 
- bulk raw counts
- pseudobulk raw counts

Much of this code will be borrowed from:
- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/bulk-deconvolution/scripts/calculate-tpm.R
- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/pseudobulk-bulk-prediction/scripts/prepare-bulk-counts.R
- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/pseudobulk-bulk-prediction/scripts/calculate-pseudobulk.R

As part of these calculations, we'll also want to convert ensembl ids -> gene symbols. In addition, we're using `tidyestimate` for analysis, which takes a wide data frame where the first column is the gene symbols. So, we should export data in this format to facilitate next steps.

sjspielman:
Closes #153 

This PR adds a script to prepare data inputs for ESTIMATE. I ended up deciding to prepare all three data versions (bulk counts, tpm, pseudobulk) in the same script so that we export 1 file per project with a list of three data frames, for each expression flavor. Each of these lists is also formatted to head directly into `tidyestimate`. This also means those directories I created but didn't need yet? ...Well, we don't need them at all in the end!

As part of this, I also had to convert ensembl ids to gene symbols and handle some duplicate symbols. I decided to just sum them up. Notably, none of the duplicate genes are in the marker gene lists, so this shouldn't have much of an effect. Let me know if you wanted to try out other approaches here, but again, I think this won't affect inference much either way.

### DIFF
--- a/analysis/estimate/run-estimate-analysis.sh
+++ b/analysis/estimate/run-estimate-analysis.sh
@@ -12,12 +12,8 @@ cd "$basedir"
 data_dir="data"
 script_dir="scripts"
 scpca_dir="${data_dir}/scpca"
-tpm_dir="${data_dir}/tpm"
-pseudobulk_dir="${data_dir}/pseudobulk"
 
 mkdir -p $scpca_dir
-mkdir -p $tpm_dir
-mkdir -p $pseudobulk_dir
 
 ensembl_symbol_map_file="${data_dir}/ensembl-symbol-map.tsv"
 
@@ -25,3 +21,16 @@ ensembl_symbol_map_file="${data_dir}/ensembl-symbol-map.tsv"
 Rscript ${script_dir}/sync-data-files.R \
   --output_dir "${scpca_dir}" \
   --ensembl_symbol_map_file "${ensembl_symbol_map_file}"
+
+for project_dir in $scpca_dir/*; do
+    project_id=$(basename $project_dir)
+    
+    expression_file="${data_dir}/${project_id}_expression.rds"
+    project_dir="${scpca_dir}/${project_id}"
+    
+    # Calculate all expression quantities 
+    Rscript ${script_dir}/prepare-expression-data.R \
+      --project_id "${project_id}" \
+      --scpca_dir "${project_dir}" \
+      --output_file "${expression_file}"
+done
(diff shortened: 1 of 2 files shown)


analysis/estimate/run-estimate-analysis.sh
```diff
@@ -12,16 +12,25 @@ cd "$basedir"
...
 # Sync data files from S3
 Rscript ${script_dir}/sync-data-files.R \
   --output_dir "${scpca_dir}" \
   --ensembl_symbol_map_file "${ensembl_symbol_map_file}"
+
+for project_dir in $scpca_dir/*; do
+    project_id=$(basename $project_dir)
+    
+    expression_file="${data_dir}/${project_id}_expression.rds"
+    project_dir="${scpca_dir}/${project_id}"
```
allyhawkins, line 29: Do you need to define this if you are already defining it above in the loop?

analysis/estimate/scripts/prepare-expression-data.R
```diff
@@ -0,0 +1,206 @@
...
+
+
+# Export RDS -----------------
+
+expr_list <- list(
+  "bulk_counts" = bulk_df, 
+  "bulk_tpm" = tpm_df, 
+  "pseudobulk" = pseudo_df
+)
+# check that all data frames are compatible:
```
allyhawkins, line 196: Is this comment meant to be here?

allyhawkins (approved):
LGTM. I don't have a strong preference about what to do regarding the duplicate matches, so summing is fine, especially if none of these genes are really being used when running ESTIMATE.

analysis/estimate/run-estimate-analysis.sh
sjspielman: sure don't!!

analysis/estimate/scripts/prepare-expression-data.R
sjspielman: It is, to preface the next lines which check that they have the same genes (at least the two bulks) and all have the same samples, but the spacing is unclear I think so I'll tweak that!

PR #163, opened 2025-02-24
Add considerations for low-quality cells

### LINKED ISSUE #162: Consider low-quality samples in estimate/consensus comparison
> These didn't make it into the last notebook, but perhaps it should make it into this notebook and we should exclude these samples from scatterplots (and correlations!) comparing between estimate and cell type fractions. Alternatively, we can make 2 versions of plots. I'd also want to point out in the cell fraction barplots which samples are low-quality. This could (should?) also be a subsequent PR too. What do you think all around here?

Yes we should exclude these in the scatterplots. But I think that might be a subsequent PR. I would merge this in and then update it with the points you mentioned above.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpca-paper-figures/issues/160#issuecomment-2679061199_

sjspielman:
Closes #162 

This PR updates the notebook comparing ESTIMATE & consensus cell types to take low quality samples into consideration. I made the following changes:

* In the barplots of cell type fractions, I box the samples which are designated low-quality
* In the scatterplots faceted by project, I show these samples with open points
* In the scatterplots with all projects combined & correlations, I _removed_ these samples entirely. The immune correlation is indeed a _bit_ higher when we remove these samples which _may_ be contributing more noise than signal. 
* I added a bit of text as well to contextualize these additions.
* Note that in the last PR, I accidentally didn't stash a bit of code I was trying out for this in the final scatterplot, so I restored that code here since I didn't use it in the end

For review: [compare-estimate-consensus.nb.html.zip](https://github.com/user-attachments/files/18947585/compare-estimate-consensus.nb.html.zip)


allyhawkins (approved):
LGTM. I don't think that the conclusions really change, but I think good to note these samples.

PR #171, opened 2025-02-25
Create geneset TSVs for ORA

### LINKED ISSUE #168: Determine gene sets for over-representation analysis
We're going to perform an over-representation analysis with permutation testing to ask if a set of genes is over-represented in _positive_ outliers - aka, over-represented in bulk compared to pseudobulk in a model `bulk ~ pseudobulk`.  This is part of the `pseudobulk-bulk-prediction` analysis and will use the linear model residuals.

We will use an odds ratio (OR) as our statistic:

```
 (in geneset & outlier) / (in geneset & not outlier) 
----------------------------------------------------
 (not in geneset & outlier) / (not in geneset & not outlier) 
```
This should translate to: The odds a geneset gene is an outlier compared to genes outside the geneset.

This issue tracks adding a script to export a TSV of the gene sets for a given project. The gene sets we will use should be taken from the tissue-specific PanglaoDB gene sets (where 1 gene set = 1 PanglaoDB cell type)  we used in `scpca-nf` for cell typing. We used one of these per project based on expected biology of that tumor type. 

We'll select these genesets with this approach:
- Identify the consensus cell types present across all samples
- Determine which PanglaoDB gene sets (cell types) of the tissue-specific PanglaoDB gene sets actually contributed to these consensus annotations. We can use this [consensus reference mapping TSV](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/3d1b268addb31396c27f3e1df3a49f33955a6872/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv) to do this.
- We'll then have a set of `N` gene sets to test per project.

sjspielman:
Closes #168 

This PR adds a new script to establish TSV files with gene sets to use for over-representation analysis. The script prepares a TSV for a single project with this approach:
- get all consensus cell types in the project
- determine their underlying panglao cell types
- get the union of unique ensembl ids for those cell types from the Panglao cellassign reference files
- those ensembl ids comprise the gene set for the associated consensus cell type label

I also updated this analysis' sync script to include a step to download the Panglao reference files.

Note that this code lives in the `analysis/pseudobulk-bulk-prediction` directory, but the consensus cell types were previously downloaded to the `analysis/estimate` directory and I didn't think it was a good use of time to port that code into this directory, so we just point to the `estimate` directory to get those files. Also while doing this, I added an argument to the analysis run script to, by default, skip the GSEA analysis since it takes like 2 hours with all the shuffling.

The files this script exports are in the `data` directory so they are currently .gitignore'd. I'm therefore attaching them here in a zip file so you can see how they look more easily.
[panglao-geneset-tsvs.zip](https://github.com/user-attachments/files/18971272/panglao-geneset-tsvs.zip)

I think this is your first time directly reviewing code in this analysis directory, so please let me know what and where I can clarify!

### DIFF
--- a/analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
+++ b/analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
@@ -1,0 +1,132 @@
+# This script prepares gene sets for input to over-representation analysis (ORA) for a given ScPCA project as follows:
+# 1. Identify all unique consensus cell types across the project 
+# 2. Identify the PanglaoDB cell types that contributed to those consensus labels
+# 3. Using the scpca-nf PanglaoDB cell type reference for the given project, determine the genes in those PanglaoDB cell types
+# 4. These (unique) genes define the gene set for a given consensus cell type to test in ORA
+
+
+renv::load()
+library(optparse)
+
+
+# Parse options --------
+option_list <- list(
+  make_option(
+    "--celltype_dir",
+    type = "character",
+    help = "Directory with consensus cell types for the given project"
+  ),
+  make_option(
+    "--panglao_geneset_file",
+    type = "character",
+    help = "Path to gene set file used to cell type the given project"
+  ), 
+  make_option(
+    "--map_file",
+    type = "character",
+    help = "Path to TSV file mapping bulk sample and library ids, which is needed because not all samples are used in all projects"
+  ),
+  make_option(
+    "--output_file",
+    type = "character",
+    help = "Path to output TSV file to save gene sets."
+  ), 
+  make_option(
+    "--consensus_reference_url",
+    type = "character",
+    default = "https://raw.githubusercontent.com/AlexsLemonade/OpenScPCA-analysis/refs/heads/main/analyses/cell-type-consensus/references/consensus-cell-type-reference.tsv",
+    help = "URL with map between consensus cell types and PanglaoDB gene set labels"
+  ), 
+  make_option(
+    "--library_metadata_file",
+    type = "character",
+    default = here::here("s3_files", "scpca-library-metadata.tsv"),
+    help = "Path to ScPCA library metadata file."
+  )
+)
+opts <- parse_args(OptionParser(option_list = option_list))
+
+# Checks---------
+stopifnot(
+  "Panglao gene set file not found" = file.exists(opts$panglao_geneset_file), 
+  "Library metadata file not found" = file.exists(opts$library_metadata_file), 
+  "ID map file not found" = file.exists(opts$map_file)
+)
+
+# Prepare input files  -----------------
+consensus_ref_df <- readr::read_tsv(opts$consensus_reference_url)
+
+panglao_df <- readr::read_tsv(opts$panglao_geneset_file) |>
+  tidyr::pivot_longer(
+    -ensembl_id, 
+    names_to = "panglao_celltype", 
+    values_to = "gene_present"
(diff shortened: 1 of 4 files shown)


sjspielman:
@allyhawkins I requested you for review here since you're the most familiar with the TSV files I'm using in this PR, but as discussed in DSTM I plan to request @jaclyn-taroni for the notebook that actually does the analysis.

analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
```diff
@@ -0,0 +1,132 @@
...
+  recursive = TRUE
+) |>
+  purrr::set_names(
+    \(x) {stringr::str_split_i(basename(x), pattern = "_", i = 1)}
+  )
+
+# determine libraries to keep from the consensus cell types
+# we need to take this step because some projects don't use all their single-cell libraries
+keep_libraries <- readr::read_tsv(opts$map_file) |>
+  dplyr::select(scpca_sample_id) |>
```
allyhawkins, line 79: I would approach this differently. 
First, filter the library metadata to the project IDs and those that are in bulk and pull the sample IDs. This gives you the list of sample IDs that have bulk RNA-seq. 
Then filter the library metadata to seq unit in cell/ nucleus and sample ID in bulk list and pull the library IDs that way. 

This should eliminate needing that extra map file.

analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
```diff
@@ -0,0 +1,132 @@
...
+
+# Combine these annotations with the consensus map to get their
+# original panglao names
+consenus_panglao_df <- consensus_ref_df |>
+  dplyr::filter(consensus_annotation %in% present_celltypes) |>
+  dplyr::select(consensus_annotation, original_panglao_name) |>
+  dplyr::distinct() 
+
+# For each consensus group, determine the (unique) panglao marker genes that fed into it.
+# This creates a table with columns `gene_set_name` and `ensembl_id`
```
allyhawkins, line 114:
```suggestion
# This creates a table with columns `cell_type_name` and `ensembl_id`
```
I think to avoid confusion we should just refer to these as cell types rather than names of gene sets.

allyhawkins (commented):
Generally this looks good, but I still think we don't need that extra map file.

analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
sjspielman: For this, I'll also need to filter out multiplexed (which can be done by removing any rows where there is a `;` in the sample id), and add an argument for project id, but yeah this should be fine!

analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
sjspielman: Given this suggestion, do you think we should also rename the script?

sjspielman: I don't think it matters much either way really, so I'll leave the script name but take the suggestion

sjspielman:
I've updated the parsing here to remove the need for the map file and capture all the samples. Note that I ran this _after_ the code in #173, so the tables I generated were made from all the samples including SCPCS000144.

PR #173, opened 2025-02-26
Add missing sample to analysis

### LINKED ISSUE #169: Add notebook for over-representation analysis
Following #168, we'll need to add the notebook for over-representation analysis itself, for which we'll do permutation tests to assess significance. 

The procedure for each project is as follows:
- Average the residuals across samples for each gene. We will then have data of "average residuals" to proceed with 
- Identify outliers in either direction: >2.5 sd and < 2.5 sd to test separately. The former tests the hypothesis of bulk overrepresentaiton, and the latter tests the hypothesis of pseudobulk overrepresentation. 
  - note the sd is from the averaged residuals - not the underlying linear model
- For each gene set, build a contingency table and calculate the OR as our data statistic
- Then, perform permutations:
  - Shuffle gene/residual relationship 1000 times
  - At each rep, calculate the same odds ratios
  - In the end we'll have 2 shuffled OR distributions for each gene set (one in each direction bulk or pseudobulk) from which we can calculate P-values

This can be implemented in a notebook with 2 or 3 parameters:
- the outlier detection threshold as the number of standard deviations from the averaged residuals (default 2.5)
- the number of reps for permutation (default 1000)
- I do think we can reasonably have all projects in 1 notebook, but if it proves really unwieldy then I will also add a project parameter so this notebook can be used as a template

Noting also that it might be a helpful viz to make residual plots and color our marker genes of interest.

### LINKED ISSUE #172: Update sync script in prediction analysis to include SCPCS000144
Currently in `analysis/pseudobulk-bulk-prediction`, we select samples for analysis in part by filtering out any samples with `cellhash` technology. I am late to the party to learn about our sample SCPCS000144 from SCPCP000009, which is indeed cellhash but there's not actually another sample multiplexed with it. Our current parsing excludes this sample, but we should keep it.

This issue tracks updating the sync script to include this sample (and take a beat to confirm there aren't any more that got lost along the way!), and then re-generate downstream analysis notebooks to include the missing sample.

sjspielman:
~Closes #169~
Closes #172!! (woops, used wrong issue in branch name, but we'll survive!)

This PR updates the original parsing code in `sync-data-files.R` to find all relevant samples for analysis so that the same we are missing, SCPCS000144, is included. Rather than finding multiplexed via technology, I now search based on presence of multiple samples per a single-cell/nucleus library id. 

I then re-ran the analysis pipeline (including TPM calculations which are no longer used, but this allowed me to render notebooks fresh). I re-rendered the notebooks:
- Two exploratory notebooks: 
- All model notebooks
- _not yet_ the GSEA notebooks, because they take a little while.. I'm going to open a separate issue to make sure we re-render those

While re-running, I also fixed a typo in the `run-prediction.sh` script where the wrong output file flag was specified.


sjspielman:
Noting I'll make sure to keep #169 open once this goes in!

allyhawkins (approved):
LGTM

PR #175, opened 2025-02-26
Add notebook and results for ORA

### LINKED ISSUE #169: Add notebook for over-representation analysis

sjspielman:
**Stacked on #168**
Closes #169 

This PR adds a notebook to perform over-representation analysis using project-specific gene sets (for cell types) established in #168.

In the end, I made this as a template notebook to run the analysis on a single project, and I export project-specific rendered HTML files to `notebooks/ora-htmls/`. I also export a TSV to `results/` of the odds ratios and their P-values following permutation testing. 

The notebook takes the following steps on a given project:
- Calculate the per-gene median residual value 
  - Note that I have an Rmd parameter here if we want to try mean, as well
- For each gene set (cell type), calculate the odds ratios in both the bulk and pseudobulk over-representation direction from the real data 
  - For this, I used a threshold of 2.5 (also an Rmd parameter) for outlier detection. I zscore'd the summarized residuals first to be able to use the 2.5 threshold directly for bulk (and similarly -2.5 for pseudobulk) 
- Create 1000 shuffled data versions and re-calculated these odds ratios on each
- Visualize the shuffled odds ratio distributions as histograms with a vertical line showing the odds ratio calculated on the real data.
- P-values are then defined as the fraction of shuffled replicates that are `>=` the odds ratio calculated on the real (not shuffled) data
  - I further multiplied these by 2 since we're doing 2 tests per gene set (cell type) - feedback here much appreciated!
- Notebooks display the significant result gene sets (cell types) 
- Finally, export table of odds ratios and p-values

Here's a summary of the significant results at 0.05, which _mostly_ came up in the **bulk** direction:

- SCPCP000001:
	- **pseudobulk** direction: `innate lymphoid cell` , `T cell` and `natural killer cell`
	- **bulk** direction: `astrocyte`, `endothelial cell`, `macrophage`, `myeloid leukocyte`, `neuron`, `neuron associated cell`, `neutrophil`
- SCPCP000002:
	-  **pseudobulk** direction: `T cell`
- SCPCP000006:
	- **bulk** direction : `regulatory T cell` and `T cell`
- SCPCP000009:
	- **pseudobulk** direction: 6`astrocyte`, `macrophage`, `monocyte`, `mononuclear phagocyte`, `myeloid leukocyte`, `neuron associated cell`
- SCPCP000017:
	- 23 cell types in **bulk** direction (not listed because 23 is a lot)

I'm attaching the HTMLs directly as well to assist in review: [ora-htmls.zip](https://github.com/user-attachments/files/18988176/ora-htmls.zip)


analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,270 @@
...
+  # (in geneset & outlier) / (in geneset & not outlier) 
+  # divided by
+  # (not in geneset & outlier) / (not in geneset & not outlier) 
+  # This is:
+  # The odds a cell type marker gene is an outlier compared to non-cell-type marker genes
+  odds_num <- (dplyr::filter(df, in_geneset, outlier) |> dplyr::pull(n)) / (dplyr::filter(df, in_geneset, !outlier) |> dplyr::pull(n))
+  odds_denom <- (dplyr::filter(df, !in_geneset, outlier) |> dplyr::pull(n)) / (dplyr::filter(df, !in_geneset, !outlier) |> dplyr::pull(n))
+  
+  return(odds_num / odds_denom)
+}
```
jaclyn-taroni, line 66: As discussed in our meeting earlier this week, ORs take everyone a second 😄 

As discussed in Slack, this is _equivalent_ to how I think about the problem (odds an outlier is in a gene set compared to non-outliers), but your way is also how @jashapiro thinks about the problem.

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,270 @@
...
+        bulk_pvalue = sum(shuffled_or_bulk >= baseline_bulk_or) / params$reps, 
+        pseudobulk_pvalue = sum(shuffled_or_pseudobulk >= baseline_pseudobulk_or) / params$reps
+      ) |>
+      # correct pvalue for 2 tests, capping it at 1
+      dplyr::mutate(
+        bulk_pvalue = min(2*bulk_pvalue, 1),
+        pseudobulk_pvalue = min(2*pseudobulk_pvalue, 1),
+      )
+    }) |>
+    purrr::list_rbind(names_to = "geneset_cell_type")
```
jaclyn-taroni, line 243: As discussed in Slack, calculate an FDR using the p-values from both "directions."

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,270 @@
...
+
+```{r}
+shuffled_or_df <- 1:params$reps |>
+  purrr::set_names(1:params$reps) |>
+  purrr::map(\(i) {
+    
+    shuffled_df <- data.frame(
+      ensembl_id = summary_residual_df$ensembl_id, 
+      summary_residual = sample(summary_residual_df$summary_residual, nrow(summary_residual_df))   
+    )
```
jaclyn-taroni, line 159: This looks right to me 👍🏻

jaclyn-taroni (commented):
Returning some "science feedback" -- most of which is just recording what folks have talked about in Slack for posterity.

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,270 @@
...
+    # Define these variables for code clarity
+    baseline_bulk_or <- geneset_baseline_df$or_bulk
+    baseline_pseudobulk_or <- geneset_baseline_df$or_pseudobulk
+      
+    shuffled_or_df |>
+      dplyr::filter(geneset_cell_type == geneset_name) |>
+      # Calculate P-value as fraction of shuffled ORs >= baseline OR
+      dplyr::summarize(
+        bulk_pvalue = sum(shuffled_or_bulk >= baseline_bulk_or) / params$reps, 
+        pseudobulk_pvalue = sum(shuffled_or_pseudobulk >= baseline_pseudobulk_or) / params$reps
```
jaclyn-taroni, line 235: Oops, meant to say this looks correct as well – as or more extreme than the actual OR

jashapiro: Somewhere in here you may also want to set a minimum pvalue of 1/reps (or 1/(reps+1), depending how pedantic you want to be). Otherwise FDR won't properly reflect corrections for "most extreme" values.

sjspielman: @jashapiro I'm not familiar with this actually? I've done comparatively less fdr correction in my day vs bonferroni or such... 
Here's what I interpret this to mean:

```
      dplyr::mutate(
        q_value = p.adjust(p_value_uncorrected, method = "BH"), 
        q_value = dplyr::case_when(
          q_value > 1 ~ 1, 
          #### this line:
          q_value < 1/(params$reps+1) ~ 1/(params$reps+1),
          .default = q_value
     ))
```
Where `params$reps` is 1000 by default. Is this the idea?

jaclyn-taroni: Good point/catch, @jashapiro. I thought about that and then promptly forgot. The point is to set the minimum p-value as Josh describes _before you correct_, @sjspielman

sjspielman: Aha, so something like this (pseudocode) where `threshold = 1(params$reps)+1)`?

```
mutate(p = ifelse(p < threshold, threshold, p), 
       q = p.adjust(p, method = "BH") # p.adjust already upper bounds at 1
)
```

jaclyn-taroni: That looks right to me.

jashapiro: > Aha, so something like this (pseudocode) where `threshold = 1/(params$reps+1)`?
> 
> ```
> mutate(p = ifelse(p < threshold, threshold, p), 
>        q = p.adjust(p, method = "BH") # p.adjust already upper bounds at 1
> )
> ```

Yes, but I would use `max(p, threshold)`

jashapiro: A bit of further explanation: it isn't about the FDR per se as much as just calculating p values from permutations. Since the real p value is never 0, you can set the minimum as if the _next_ permutation would have passed whatever threshold. To be more exact/conservative, you could apply this always, to set all p values as  (n_extreme + 1)/(n_perm + 1).

sjspielman:
I've implemented the feedback on p-value corrections, and added an intro section that includes a more thorough explanation of the odds ratio.
With the FDR as implemented, we no longer have any significant results. Here is one of the rendered notebooks for review: 
[perform-ora_SCPCP000001.nb.html.zip](https://github.com/user-attachments/files/19014654/perform-ora_SCPCP000001.nb.html.zip)

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+      # Correct for multiple testing with FDR
+      tidyr::pivot_longer(
+        everything(),
+        names_to = "modality", 
+        values_to = "p_value"
+      ) |>
+      dplyr::mutate(
+        # first ensure p-value is at least 1/(params$reps+1)
+        p_value = max(p_value, 1/(params$reps + 1)),
+        q_value = p.adjust(p_value, method = "BH")
```
jaclyn-taroni, line 276: I know the term FDR adjusted p-value and q-value are used interchangeably in practice, but they are not exactly the same:
```suggestion
        p_adj_bh = p.adjust(p_value, method = "BH")
```

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+  }) |>
+  purrr::list_rbind(names_to = "geneset_cell_type")
+```
+
+
+How many gene sets (cell types) are significant per modality, considering the q-value (FDR corrected p-value)?
+
+```{r}
+permutation_pvalues |>
+  dplyr::filter(q_value <= 0.05) |>
```
jaclyn-taroni, line 287:
```suggestion
  dplyr::filter(p_adj_bh <= 0.05) |>
```

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+
+What are those gene sets (cell types), if any?
+
+```{r}
+# wrangle a bit to make this easily human-readable
+permutation_pvalues |>
+  dplyr::select(-p_value) |>
+  dplyr::filter(q_value <= 0.05) |>
+  dplyr::group_by(modality) |>
+  dplyr::arrange(q_value, .by_group = TRUE)
```
jaclyn-taroni, line 299:
```suggestion
  dplyr::filter(p_adj_bh <= 0.05) |>
  dplyr::group_by(modality) |>
  dplyr::arrange(p_adj_bh, .by_group = TRUE)
```

jaclyn-taroni (approved):
One quibble with a column name, but LGTM otherwise

jashapiro (commented):
I left a number of comments, but the big one is at the end. The results in the table (if that is up to date) don't make sense to me in light of the plots that you have. For the very first cell type (astrocytes) in the first sample (01), the p-value looks like it should be small, but in the table is shown as 1. So something is going wrong somewhere along the line. The directionality you chose looks right, but since no shuffled data seems to approach a value of 4, I would expect the p value to be ~0.001.

image

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+      # Correct for multiple testing with FDR
+      tidyr::pivot_longer(
+        everything(),
+        names_to = "modality", 
+        values_to = "p_value"
+      ) |>
+      dplyr::mutate(
+        # first ensure p-value is at least 1/(params$reps+1)
+        p_value = max(p_value, 1/(params$reps + 1)),
+        p_adj_bh = p.adjust(p_value, method = "BH")
```
jashapiro, line 276: Ah, here is the error: this would need to be done `rowwise` or you just get the max p. Probably better to just calculate the p value by the more conservative formula of `(n_extreme + 1)/(n_reps +1) from the start.

sjspielman: The number of times I have been gotten by needing a `rowwise` is non-zero.

jashapiro: or `pmax()` in this case.

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+
+We use an odds ratio as our test statistic from this contingency table:
+
+```markdown
+|---------------------------------| Gene is outlier | Gene is not outlier |
+|---------------------------------|-----------------|---------------------|
+|Gene is cell type marker gene    |       A         |       B             |
+|Gene is not cell type marker gene|       C         |       D             |
+|---------------------------------|-----------------|---------------------|
+```
```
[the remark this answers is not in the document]
sjspielman: The honest answer is that this started out as a comment before I moved it to the intro. This does make more sense!

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+  dplyr::count(modality)
+```
+
+What are those gene sets (cell types), if any?
+
+```{r}
+# wrangle a bit to make this easily human-readable
+permutation_pvalues |>
+  dplyr::select(-p_value) |>
+  dplyr::filter(p_adj_bh <= 0.05) |>
```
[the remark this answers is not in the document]
sjspielman: Just for the record, this isn't actually removing it from being _saved_ - just for looking at the table. Why did I put it there, you asked? For me, it was for my eyeballs to look at the right column!

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+  dplyr::summarize(
+    summary_residual = ifelse(
+      params$summary_function == "mean",
+      mean(residual), 
+      median(residual)
+    ) 
+  ) |>
+  # zscore for outlier detection, but have to name separately due to how scale() works
+  dplyr::mutate(scale(summary_residual)) |>
+  dplyr::rename(scaled_summary_residual = `scale(summary_residual)`)
```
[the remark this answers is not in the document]
sjspielman: I'm seeing this for column naming with e.g. `dplyr::mutate(x = scale(summary_residual)) |>`:

```
# A tibble: 45,326 × 3
   ensembl_id      summary_residual   x[,1]
   <chr>                      <dbl>   <dbl>
 1 ENSG00000000003            2.97   1.17  
 2 ENSG00000000005           -0.295 -0.108 
 3 ENSG00000000419            0.477  0.195 
 4 ENSG00000000457            0.942  0.377 
 5 ENSG00000000460            0.147  0.0655
 6 ENSG00000000938            1.63   0.647 
 7 ENSG00000000971            5.47   2.15  
 8 ENSG00000001036            2.23   0.882 
 9 ENSG00000001084            1.27   0.504 
10 ENSG00000001167            2.64   1.04 
```

Playing with this a bit, I realize that it behaves just fine if you straight up refer to it as `x`... Hm. The display is a lie.

sjspielman:
PSA, I've just caught a big bug that we all missed 😬 

First I scaled the residuals
https://github.com/AlexsLemonade/scpca-paper-figures/blob/b5a59a0a8a82f57873f0444754c4308256d8b539/analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd#L157-L158

And then I proceeded to not use them in all manner of locations.

https://github.com/AlexsLemonade/scpca-paper-figures/blob/b5a59a0a8a82f57873f0444754c4308256d8b539/analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd#L103

https://github.com/AlexsLemonade/scpca-paper-figures/blob/b5a59a0a8a82f57873f0444754c4308256d8b539/analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd#L183-L186


I'll be using them next!

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,320 @@
...
+      tidyr::pivot_longer(
+        everything(),
+        names_to = "modality", 
+        values_to = "p_value"
+      ) |>
+      dplyr::mutate(
+        # first ensure p-value is at least 1/(params$reps+1)
+        p_value = max(p_value, 1/(params$reps + 1)),
+        p_adj_bh = p.adjust(p_value, method = "BH")
+      )
```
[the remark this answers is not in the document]
sjspielman: @jashapiro did you mean to _wholesale_ delete the `p.adjust()` line here? My assumption is no, but maybe I'm not exactly following this?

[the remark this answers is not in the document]
jashapiro: No I did not; just p_value part. oops

sjspielman:
I'm running this through now with 10,000 replicates which of course takes a bit longer but definitely not prohibitively! Project 1 is finished and has a three significant cell types over-represented in the `bulk` direction, which seem to agree with the histograms (the binwidth could be better here, honestly, but I'd rather run these through to get results than stop in the middle right now - can always rerun later). 

I'm going to re-request review at this point while the rest are still running.

[perform-ora_SCPCP000001.nb.html.zip](https://github.com/user-attachments/files/19016555/perform-ora_SCPCP000001.nb.html.zip)

sjspielman:
All notebooks are now done, and we have some significant bulk in all of them:
[ora-htmls.zip](https://github.com/user-attachments/files/19018230/ora-htmls.zip)

I didn't time this specifically, but I'd estimate it took little over an hour to run all of these.

analysis/pseudobulk-bulk-prediction/results/SCPCP000002_ORA-odds-ratios.tsv
```diff
@@ -0,0 +1,53 @@
+geneset_cell_type	modality	p_value	p_adj_bh	or_bulk	or_pseudobulk
```
jaclyn-taroni, line 1: Why do we need both `or_bulk` and `or_pseudobulk` for every row here rather than `odds_ratio` and let `modality` tell you which one it is?

sjspielman: hm, yes let me check this pivoting.

sjspielman: Noting I updated this to make it long all around in https://github.com/AlexsLemonade/scpca-paper-figures/pull/175/commits/97bd09964fd19d344cbdc3151e4b301f6d1380f1, but plan to wait to re-render notebooks in case there are more reviews.

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,311 @@
...
+  or_bulk <- residual_df |>
+    dplyr::mutate(
+      outlier = scaled_summary_residual > outlier_threshold,
+      in_geneset = ensembl_id %in% geneset_ensembl_ids
+    ) |>
+    dplyr::summarize(
+      odds_num = sum(in_geneset & outlier) / sum(in_geneset & !outlier),
+      odds_denom = sum(!in_geneset & outlier) / sum(!in_geneset & !outlier),
+      odds_ratio = odds_num / odds_denom
+    ) |>
```
jaclyn-taroni, line 95: I find this a lot clearer!

sjspielman: _much_!! thanks, @jashapiro!

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,311 @@
...
+    geneset_baseline_df <- baseline_or_df |>
+      dplyr::filter(geneset_cell_type == geneset_name)
+    
+    # Define these variables for code clarity
+    baseline_bulk_or <- geneset_baseline_df$or_bulk
+    baseline_pseudobulk_or <- geneset_baseline_df$or_pseudobulk
+      
+    shuffled_or_df |>
+      dplyr::filter(geneset_cell_type == geneset_name) |>
+      # Calculate P-value as fraction of shuffled ORs >= baseline OR?
```
jashapiro, line 246:
```suggestion
      # Calculate P-value as fraction of shuffled ORs >= baseline OR (with +1 correction on numerator & denominator to avoid p=0)
```

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,311 @@
...
+  purrr::list_rbind(names_to = "geneset_cell_type") 
+
+# Join the odds ratios in with their pvalues, all in long format
+permutation_results <- baseline_or_df |>
+  tidyr::pivot_longer(
+    starts_with("or_"), 
+    names_to = "modality", 
+    values_to = "odds_ratio"
+  ) |>
+  dplyr::mutate(modality = stringr::str_remove(modality, "or_")) |>
```
jashapiro, line 268: I don't think it matters, but to be safe, remove only the start of the string.
```suggestion
  dplyr::mutate(modality = stringr::str_remove(modality, "^or_")) |>
```

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
```diff
@@ -0,0 +1,311 @@
...
+model_dir <- here::here(result_dir, "models")
+data_dir <- here::here("analysis", "pseudobulk-bulk-prediction", "data")
+
+model_data_file <- file.path(
+  model_dir, 
+  glue::glue("{params$project_id}_bulk-pseudobulk-model_threshold-0.rds")
+) 
+geneset_file <- file.path(
+  data_dir, 
+  glue::glue("{params$project_id}_panglao-genesets.tsv")
```
jashapiro, line 69: Just for my clarification; these gene sets only include cells that were called in that project, or they are the full set of "possible" gene sets?

jashapiro (approved):
This looks good to me. 

The one thing I am a bit curious about is that I'm sort of surprised by the number of OR==0 in these tables. If I think about it it can kind of make sense, especially if the gene sets and # of outliers are small. It might be helpful for intuition to report somewhere on  the gene set sizes for each project, as power to detect OR deviations is going to be dependent in part on that. But that does not need to be part of this notebook.

sjspielman:
> It might be helpful for intuition to report somewhere on the gene set sizes for each project, as power to detect OR deviations is going to be dependent in part on that. But that does not need to be part of this notebook.

Thinking ahead, maybe we put the sizes of each gene set directly into the labels when we plot.

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
sjspielman: These only include the consensus cells that were called in that project. 
From there, it includes all marker genes for each cell type that feeds into each consensus cell type.

sjspielman:
Now that this has been approved, I am going to re-generate the notebooks with the most recent `Rmd` version, and commit the updated notebooks & TSVs. Then we can merge.

analysis/pseudobulk-bulk-prediction/notebooks/perform-ora.Rmd
jashapiro: Hmmm... I wonder if we might want to include more. Imagine that there is a cell type that is _never_ found in single cell preps from a tissue, but is present in the bulk. We'd want to know about that! What we have is interestingly more conservative and more leniant:
- more conservative in that we are giving pseudobulk maximum advantage here; we won't find what we don't look at.
- more lenient in that if we include more cell types, we are doing more tests, so the p adjustment will be more severe.

I'm not really sure what the right answer is here; I think the first part of the question I have is how many more gene sets would we be using if we used the same set of gene sets we fed into CellAssign? If it isn't too different, we might want to just use that?

sjspielman: > I'm not really sure what the right answer is here; I think the first part of the question I have is how many more gene sets would we be using if we used the same set of gene sets we fed into CellAssign? If it isn't too different, we might want to just use that?

Yeah, this is an interesting point.  I do think we'd still want to keep the gene sets fed into overrepresentation at the level of consensus cell types, not CellAssign cell types - otherwise the number of gene sets gets real big real fast, and even less independent from one another. I'm not quite familiar enough with the cell types to know off the top of my head if all cell types that comprise a consensus cell type are always present together in a given CellAssign reference. I would assume yes, but this is worth confirming. If so, that means we can collapse a given CellAssign reference into consensus cell types. It's possible that the gene sets we're already using effectively already are doing this though, so the first step here is indeed some counting!

allyhawkins: I think this is a good point, and I would imagine given how we assigned the consensus cell types that all the subtypes that comprise a consensus type are in the same reference. If they aren't then we probably have a bigger problem... 

I think you could start by just removing this line and seeing how the total number of cell types/ gene sets changes for each reference. https://github.com/AlexsLemonade/scpca-paper-figures/blob/76e51f212f32116b250f9531d8ef59157a5378cf/analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R#L115

sjspielman: So, I think they aren't? I'm going to open an issue where I looked into this so you all can see my code there without cluttering up here. Stay tuned shortly.

allyhawkins: Thinking about this more I think this point shouldn't matter. Either way you are still going to constrict the possible cell types to those that are in the panglao reference for that organ. So ultimately the consensus cell types included will be the same regardless if all or some of the panglao labels that make up that consensus cell type is present or not.

sjspielman:
Noting that here was the output from `time`:

```
real	6175.21s
user	5823.25s
sys	336.20s
```

So, a bit under _two_ hours.

PR #185, opened 2025-02-28
Add script to sync consensus celltype results 

allyhawkins:
In preparation to add figures that use the consensus cell type results, here I'm adding a script to the `figure_setup` folder that copies the output from consensus cell typing present in the workflow results bucket from OpenScPCA. 

We already had a script to sync data from S3, but I chose to do this separately since it uses openscpca results and we need an AWS profile. It's also a lot of files so it seemed like it deserved it's own script.

### DIFF
--- a/scripts/README.md
+++ b/scripts/README.md
@@ -35,6 +35,15 @@ To run the script use the following command:
 op run -- Rscript sync-reference-files.R
 ```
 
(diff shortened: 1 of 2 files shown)


sjspielman (approved):
LGTM!

PR #192, opened 2025-03-03
Remove Fig S6 and S7

### LINKED ISSUE #178: Condense supplemental figures for cell typing
Currently we have 4 supplemental figures that show various aspects of cell typing, which is a lot. We also are planning to add up to 2 more supplemental figures for the consensus cell types. The consensus cell types are arguably more interesting here, so we should condense the existing cell type figures. After going through the figures, I think we should make the following changes: 

- S4 Justification of reference choice for `SingleR`: Keep as is
- S5 Submitter-provided annotations: Keep the UMAP in panel A and replace panel B with what is currently S7 showing the comparison between submitter provided annotations, SingleR, and CellAssign 
- Remove S6 since those are all just examples from reports
- Remove S7 since that will now be part of S5

allyhawkins:
Closes #178 

Here I am condensing the supplemental figures that we have for cell type annotation. After looking at this some more and re-reading the text we have around cell typing in the results section I ended up just taking out what is currently S6 and S7. 

Figure S4 is the justification for choosing the `SingleR` reference, so we should keep that. And then Figure S5 is some justification around why we chose `CellAssign`, in particular we highlight that tumor cells (as annotated by a submitter) tend to line up with Unknown. So I do think we want to keep that since we are pointing out a particular feature of `CellAssign` that we used to choose it as a method. These are both described in the [Choosing references section of the results](https://github.com/AlexsLemonade/ScPCA-manuscript/blob/main/content/03.results.md#choosing-cell-typing-methods-and-references). I thought about condensing these into one figure, but these are pretty large panels, so for now they are separate, which I think is fine. 

Figure S6 and S7 both are showing example plots that can be found in the cell type annotation summary reports. These are both described in  the [Adding cell types to the Portal section of the results](https://github.com/AlexsLemonade/ScPCA-manuscript/blob/main/content/03.results.md#adding-cell-type-annotations-to-the-scpca-portal). Since these are just the same plots we show in the reports, I don't think they are necessary. So I took these out. 

- I removed the compiled images for these plots. 
- I moved the scripts that we used to create these plots to a new folder `scripts/old_figures`. I could just remove the scripts entirely, but I felt like I didn't want to totally get rid of that code yet? What do we think about leaving it? 
- I also removed mention of these figures from the outline in the readme and they are no longer generated with `generate-figures-tables.sh`. 

While trying to get my changes committed, I was having some issues with `pre-commit` showing errors with `R.utils`. I noticed we were using an out of date version of `pre-commit`, so I updated to the most recent version.

### DIFF
--- a/generate-figures-tables.sh
+++ b/generate-figures-tables.sh
@@ -73,8 +73,11 @@ Rscript ${script_dir}/Fig3D_merged-umaps.R
 # Figure S4
 Rscript ${script_dir}/FigS4_celldex-ref-comparison.R
 
-# Figure S5A-B
-Rscript ${script_dir}/FigS5A-B_cellassign-justification.R
+# Figure S5A
+Rscript ${script_dir}/FigS5A_cellassign-justification.R
+
+# Figure S5B
+Rscript ${script_dir}/FigS5B_submitter-celltypes-heatmap.R
 
 ##########################################################
 #         Generate tables in order of appearance         #


scripts/FigS5A-B_cellassign-justification.R
```diff
@@ -63,7 +63,7 @@ celltype_df <- celltype_df |>
                   forcats::fct_lump_n(5, other_level = "All remaining cell types", ties.method = "first") |>
                   forcats::fct_infreq() |>
                   forcats::fct_relevel("Unknown cell type", "All remaining cell types", after = Inf)
-  )
+  )
```
allyhawkins, line 66:
```suggestion
  )
```

scripts/README.md
```diff
@@ -84,8 +84,12 @@ Before running this script, you must run both `figure_setup/sync-data-files.R` a
...
-13. `FigS6A-B_celltype-diagnostic-plots.R`: This script is used to generate supplemental Figure 4A-B, which includes examples of the diagnostic plots for annotating cell types with `SingleR` and `CellAssign`.
+## Old figures
+
+The `old_figures` folder contains the scripts used to generate figures present in previous versions of the manuscript. 
+
+1. `FigS6A-B_celltype-diagnostic-plots.R`: This script is used to generate supplemental Figure 4A-B, which includes examples of the diagnostic plots for annotating cell types with `SingleR` and `CellAssign`.
 Before running this script, you must run `figure_setup/sync-data-files.R`.
 
-14. `FigS7_submitter-celltypes-heatmap.R`: This script is used to generate supplemental Figure 5, which includes an example heatmap comparing submitter provided annotations to automated annotations from `SingleR` and `CellAssign`.
+2. `FigS7_submitter-celltypes-heatmap.R`: This script is used to generate supplemental Figure 5, which includes an example heatmap comparing submitter provided annotations to automated annotations from `SingleR` and `CellAssign`.
```
jaclyn-taroni, line 94: Probably want to rewrite this so you're not describing outdated figures but are just saying what kinds of plots the scripts make.

scripts/README.md
```diff
@@ -84,8 +84,12 @@ Before running this script, you must run both `figure_setup/sync-data-files.R` a
 12. `FigS5A-B_cellassign-justification.R`: This script is used to generate supplemental Figure 7A-B, which includes a UMAP summarizing cell type annotations using `CellAssign` and a heatmap comparing the `CellAssign` annotations to submitter-provided annotations.
 Before running this script, you must run `figure_setup/sync-data-files.R`.
 
-13. `FigS6A-B_celltype-diagnostic-plots.R`: This script is used to generate supplemental Figure 4A-B, which includes examples of the diagnostic plots for annotating cell types with `SingleR` and `CellAssign`.
+## Old figures
+
+The `old_figures` folder contains the scripts used to generate figures present in previous versions of the manuscript.
```
jaclyn-taroni, line 89: This seems like a fine approach 👍🏻

figures/pdfs/FigS7_submitter-heatmap.pdf
jaclyn-taroni, line 1: Was this panel going to become part of S5? From https://github.com/AlexsLemonade/scpca-paper-figures/issues/178#issue-2887719690:


> S5 Submitter-provided annotations: Keep the UMAP in panel A and replace panel B with what is currently S7 showing the comparison between submitter provided annotations, SingleR, and CellAssign
> Remove S7 since that will now be part of S5


So then I'd maybe expect the S7 script to get renamed or consolidated with something else.

jaclyn-taroni (commented):
The big thing is that I didn't expect S7 to get deleted so much as renumbered based on #178.

allyhawkins:
> Was this panel going to become part of S5? From [#178 (comment)](https://github.com/AlexsLemonade/scpca-paper-figures/issues/178#issue-2887719690):
> 
> > S5 Submitter-provided annotations: Keep the UMAP in panel A and replace panel B with what is currently S7 showing the comparison between submitter provided annotations, SingleR, and CellAssign
> > Remove S7 since that will now be part of S5
> 
> So then I'd maybe expect the S7 script to get renamed or consolidated with something else.

Ah yes, when I was going through this in more detail while working on the PR, I had decided to not make that change, but thinking about it more, I think we do want to make the change. So I updated this to replace the previous S5B with what was S7. It's basically the same figure but shows both CellAssign and SingleR annotations compared to submitter annotations, rather than just CellAssign. You should now see the re-compiled S5 in this PR. 

Previously we had been making S5A and S5B in the same script, so I ended up taking out the S5B items from that script and making a new script that lives in the `old_figures` folder. And then I renamed the S7 script to now be S5B. 

So, to summarize these changes should be present: 

- remove old S5B
- remove S6
- move old S7 to new S5B

I believe I caught everything that needs to change related to this and this should be ready for another look @jaclyn-taroni.

PR #196, opened 2025-03-10
Export model data as separate TSV

### LINKED ISSUE #186: Update prediction analysis to use consensus cell types in s3_files
The code in `analysis/pseudobulk-bulk-prediction` currently has a script to sync consensus cell types and save them locally in that directory. However, in #185 a script is being added to add these consensus cell types to the `s3_files/` top-level directory for use throughout the repo. 

This issue tracks removing the analysis script to sync these files to instead take advantage of the one added in #185, and update paths to these files in code accordingly.

### LINKED ISSUE #188: Prepare data for figure 6
In figure 6 (#187), we plan to create manuscript figures for the bulk analysis. Data for these plots is currently in `analysis/pseudobulk-bulk-prediction`. We want to ensure that data is easily discoverable for eventual data sharing via Zenodo or similar.

The data that goes into these panels is as follows:

- Panel A: the data used to build the `threshold-0` models. 
  - Currently this is stored in an RDS file in `results/models/` along with the actual fitted models themselves.
  - We should update this code so that models and their underlying data that goes into the models are exported separately, such that the data is a standalone TSV. We can also include the residuals in this data frame too, for easier feeding into downstream notebooks. This output should be sent still to `results`.
- Panel B: the output TSV from the overrepresentation notebook
  - Currently these are TSV files in the analysis `results` directory. This is already in a good format for use and data sharing.

Therefore, the scope of this issue entails updating data management for the data that feeds into Panel A as described above. While here, it might also be helpful to add some more organization into the `results` directory, either by project or by file type. TBD.

sjspielman:
Closes #186 
Closes #188 

This PR does two main things:

- For #186, I updated the path to the consensus cell types results in `run-prediction.sh`.
- For #188, I updated the model-building notebook to _separately_ export a) the model as RDS, and b) the associated data (with residuals!) as TSV. This makes data more discoverable and usable in a future data sharing context. 
  - I also re-rendered the model notebooks to regenerate data files; works as expected! The TSVs are remain gitignored due to size


analysis/pseudobulk-bulk-prediction/.gitignore
```diff
@@ -2,6 +2,7 @@
 /data/*
 !/data/.gitkeep
 
-# Ignore RDS files in the results/models directory (but keep it present)
+# Ignore RDS and TSV files in the results/models directory (but keep it present)
 /results/models/*rds
+/results/models/*tsv
 !/results/models/.gitkeep
```
sjspielman, line 8:
```suggestion
!/results/models/.gitkeep

```

allyhawkins (approved):
LGTM

PR #197, opened 2025-03-10
Update the gene set script

### LINKED ISSUE #189: Update script to prepare gene sets
Broken out of #184 

This issue tracks updating our script to prepare gene sets: https://github.com/AlexsLemonade/scpca-paper-figures/blob/0345db24532230ae17e42928752a37990adabb8f/analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R

We would like this script to identify and export gene sets based on two different approaches:
1. The approach it currently uses
2. The "could use" approach outlined here: https://github.com/AlexsLemonade/scpca-paper-figures/issues/184#issuecomment-2694877940

We can export this in a single TSV with an indicator column for which framework was used.

sjspielman:
Closes #189 
Stacked on #196 

This PR updates the script to prepare gene sets for ORA. Previously, we considered only those consensus cell types which were actually observed in the single-cell data. Now, we'd like to potentially consider all consensus cell types which _could be_ observed. The previous approach is therefore a _subset_ of this new approach.

Therefore, I updated the script to collect all possible consensus cell type gene sets, and then I add an indicator column for whether that gene set was indeed observed in the consensus labels. The TSV files now contain three columns: `cell_type_name`, `ensembl_id`, and the new column `observed_in_singlecell` which is T/F. 

One question I have is, these gene set files are currently in `data` which is ignored by git, but these files (as well as some other TSV files in there) are indeed small enough to be under version control. Do we want to override the ignore for these TSVs? Or, might we want to save these TSVs elsewhere like results?

Either way for assistance in review, here are the gene set files I have generated here. 
[geneset_files.zip](https://github.com/user-attachments/files/19169656/geneset_files.zip)

### DIFF
--- a/analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
+++ b/analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R
@@ -124,7 +124,8 @@ consensus_cell_types <- consensus_ref_df |>
   dplyr::select(consensus_annotation, original_panglao_name) |>
   dplyr::distinct()
 
-# Create the consensus gene sets 
+# For each consensus group, determine the (unique) panglao marker genes that fed into it.
+# This creates a table with columns `cell_type_name`, `ensembl_id`, and `observed_in_singlecell`
 consensus_genesets_df <- split(consensus_cell_types, consensus_cell_types$consensus_annotation) |>
   purrr::map(
     \(df) {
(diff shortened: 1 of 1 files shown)


allyhawkins (commented):
The code looks good here, but I agree that we might want to include the gene sets in the repo.

sjspielman:
I've updated this to now keep the gene sets under version control, and I added a README for these files too.

allyhawkins (approved):
LGTM 🚀

Issue #195, opened 2025-03-10
Supplemental figure for non-brain bulk results


sjspielman:
We plan to show only the bulk ORA results from brain projects (1, 2, 9) in the main text (see #187). We will likely also want a supplementary figure that shows results from the other projects analyzed, 6 and 17. Given https://github.com/AlexsLemonade/scpca-paper-figures/issues/183, this might be figure S8, or we might need to make these panels if we've crossed our limit for supp figures.

Issue #187, opened 2025-03-03
Figure 6: add panels for results from bulk prediction


sjspielman:
We plan to create a figure 6 with two panels, A & B, as sketched out below:

**Edit: there should only be 3 panels per row for projects 1, 2, & 9**

![Image](https://github.com/user-attachments/assets/6d0a2181-78e6-42aa-aad5-e661dcc63d95)

- Each panel is a project, where rows go in order for projects 1, 2, 6 ~, 9, 17~ so projects are stacked in the final figure
  - Facet labels should include `(N = #)` for the number of samples which were used in the given project
  - This is fewer than the number of samples in project overall since low-quality samples were removed, and not all samples are used for project 9 which is multiplexed 
- Panel A
  - Binned (e.g. `geom_hex()` or `geom_bin_2d()`) scatterplot of `bulk ~ pseudobulk`, specifically considering the model we used for analysis; this is the one with threshold of 0.
  - Plots should include the overall regression line
  - It is likely that oob squish will come in handy here again
- Panel B
  - Barplot of odds ratios for significant gene sets, filled by -log10p
  - In the x-axis gene set labels, we should consider including `(N = #)` for the number of cell types which were included in each gene set

Since this figure is entirely ggplot with no other plotting engines, I wonder if we should just use `patchwork` from the get-go and not compile panels by hand? We can decide this when we get to coding.

This should be accomplished in 2 PRs (even if same script):

- [ ] Create panel 6A
- [ ] Create panel 6B

allyhawkins:
Just for my understanding, I thought we were only showing the brain projects in the main figures and then will have the non-brain tumors in the supplemental?

jaclyn-taroni:
> Just for my understanding, I thought we were only showing the brain projects in the main figures and then will have the non-brain tumors in the supplemental?

That was also my takeaway from the discussion on Friday.

sjspielman:
Thanks folks! It's _indeed_ very much in my notes. Very much not in my sketch.

sjspielman:
Note issue #200 as well - this code should include a table export of helper counts.

PR #202, opened 2025-03-12
Add table with bulk analysis manuscript numbers

### LINKED ISSUE #200: Add helper table with bulk analysis information
To write the manuscript sections for the bulk prediction analysis, we will need to tabulate some values. We have previously created convenience tables for this to ensure accuracy, so we should do the same here.

It will be useful to have at least these pieces of information:
- Number of samples identified with paired bulk/single-cell
- Number of samples per project which were analyzed
  - Several samples were removed for low quality, so this differs from the first bullet. 
- Number of gene sets used per project in the over-representation analysis

We can have a wide table with one row per project and these columns:
- `n_samples_total`
- `n_samples_low_quality`
- `n_samples_used` (this is just `n_samples_total` - `n_samples_low_quality`)
- `num_genesets`

The table with this information should be exported to `manuscript-numbers`. To match how we have generated tables previously, the code to actually perform this can be included with the script that creates the bulk prediction figures (#187, #195).

sjspielman:
Closes #200 

This PR initiates the figure script that will be written for #187 & #195, but currently only takes care of the task in #200. The script does not yet make figures, but it includes a section to create and export a table of counts that will be useful for manuscript writing. This includes per-project numbers of how many samples and gene sets were used, as a well as a row which gives the total counts for convenience. These numbers are included:
- `n_samples_total`: The number of samples identified as potentially suitable for analysis
- `n_samples_low_quality`: The number of samples removed due to low quality
- `n_samples_used`:  This is just `n_samples_total - n_samples_low_quality`, and is the actual number of samples which fed into the analysis in the end
- `n_ora_genesets`: The number of gene sets considered in overrepresentation analysis for each project. For this row, I didn't include a sum since that number isn't particularly meaningful.

I think this is basically all we'd need to include in text descriptions, but if/when it turns out we need additional numbers, we can always modify this code! But, if you can think of any more now that I haven't thought of let me know!

### DIFF
--- a/manuscript-numbers/bulk-analysis-counts.tsv
+++ b/manuscript-numbers/bulk-analysis-counts.tsv
@@ -1,0 +1,7 @@
+scpca_project_id	n_samples_total	n_samples_low_quality	n_samples_used	n_ora_genesets
+SCPCP000001	18	1	17	36
+SCPCP000002	20	2	18	36
+SCPCP000006	39	4	35	31
+SCPCP000009	4	0	4	36
+SCPCP000017	24	1	23	37
+total	105	8	97	NA
--- a/scripts/Fig6-FigS7_bulk-analysis.R
+++ b/scripts/Fig6-FigS7_bulk-analysis.R
@@ -1,0 +1,99 @@
+# Generates figures and manuscript numbers tables for bulk analysis
+
+renv::load()
+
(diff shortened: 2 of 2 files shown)


allyhawkins (approved):
LGTM

PR #207, opened 2025-03-13
Update dotplot widths

### LINKED ISSUE #206: Make dotplots a bit wider
I was looking at the newly added dotplot figures and noticed that a lot of the dots are overlapping, for example as screenshotted from the 5A PDF but other panels have this issue too:

Image

I'm betting making the exported figure size wider than 23 would fix this?
https://github.com/AlexsLemonade/scpca-paper-figures/blob/58e58db1d837ea7468e858c41fdde840a765f561/scripts/Fig5A-S6_consensus-cell-type-dotplots.R#L100

sjspielman:
Closes #206 

This PR updates dotplots so the dots no longer overlap. The main consequence to this is plots are no longer all the same width, which matters _somewhat_ for aligning panels in the supp figure. I thought that non-overlapping dots was probably more important than panel width for "only" the supp though.

Let me know if we're ok with this updated width strategy!


allyhawkins (approved):
I had thought about doing this when I was making these plots, but was worried how that would affect stacking them up into one figure. I agree these look better individually so I'm going to approve with the caveat that if it looks bad when compiled we should go back to using the same width.

Issue #182, opened 2025-02-28
Figure 5C: Stacked bar chart for brain/CNS tumors with only immune cells


allyhawkins:
Panel 5C will be another stacked bar chart for the brain/CNS tumors, but this time just showing immune cells. Here we want to emphasize the T cell vs myeloid cell types and will color cell types based on their original consensus cell type annotation (not the validation group). We should pick similar shades for all T cells and similar shades of a contrasting color for all myeloid cell types.

allyhawkins:
@sjspielman I was envisioning focusing this plot on the two immune cell types that we see differences in between HGG and LGG, so only coloring T cell types and myeloid types. The other cell types aren't as prevalent across these tumors, so I think it would be more visually appealing if we pick shades of one color for T cells and shades of another color for myeloid and then all other immune cells would be a grey color and grouped together. 

It probably makes sense to use shades of blue for T cell and shades of green for all myeloid subtypes based on what we have in the dot plot.

sjspielman:
> I was envisioning focusing this plot on the two immune cell types that we see differences in between HGG and LGG, so only coloring T cell types and myeloid types. The other cell types aren't as prevalent across these tumors, so I think it would be more visually appealing if we pick shades of one color for T cells and shades of another color for myeloid and then all other immune cells would be a grey color and grouped together.

Aha, this is an important nuance I missed, thank you!!

sjspielman:
@allyhawkins to confirm from this list I compiled (assuming I got all the relevant projects), we're looking at this breakdown. 

- T cell types: memory T cell, mature T cell, T cell, regulatory T cell, CD4-positive alpha-beta T cell, mature alpha-beta T cell
- Myeloid types: macrophage, mononuclear phagocyte, monocyte, myeloid leukocyte, neutrophil, dendritic cell, granulocyte
- Not included in the figure: any of the B cells, natural killer cell, plasma cell, innate lymphoid cell

Is this what you had in mind? I also think for any of these cell types that are literally represented in the validation figure (e.g. macrophage, monocyte etc), we should use the same color. Otherwise, we'll go with shades of the blue and green respectively for T and myeloid.

allyhawkins:
> [@allyhawkins](https://github.com/allyhawkins) to confirm from this list I compiled (assuming I got all the relevant projects), we're looking at this breakdown.
> 
[...]

Issue #181, opened 2025-02-28
Supplemental Figure with stacked bar charts for all diagnosis groups


allyhawkins:
We want a supplemental figure with a stacked bar chart showing all cells, including unknown cell types for all samples in a given diagnosis group. For colors and labels, we will use the `validation_group_annotation` as found in https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-validation-groups.tsv.  This should be three panels, each diagnosis group that is not the brain tumors since that is in the main figures. 

See https://github.com/AlexsLemonade/scpca-paper-figures/issues/161#issuecomment-2682654925 for an example of what the chart should look like.

PR #208, opened 2025-03-13
Add bar chart for HGG vs. LGG showing all celltypes

### LINKED ISSUE #180: Figure 5B: Stacked bar chart showing consensus cell type assignments for all cells
Figure 5B will be the stacked bar chart showing all cells, including unknown cell types. For colors and labels, we will use the `validation_group_annotation` as found in https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-validation-groups.tsv.  We will also facet by HGG vs. LGG. 

See https://github.com/AlexsLemonade/scpca-paper-figures/issues/161#issuecomment-2682654925 for an example of what the chart should look like.

allyhawkins:
Closes #180 

Here I am adding the code and the plot for Figure 5B which shows the stacked bar chart for all HGG and LGG samples that are not-multiplexed. Each bar is filled with the broad celltype group and matches up with the same colors used in the dot plot. 

- I created a new TSV file `brain-classifications.tsv` that takes all the diagnoses from the "Brain and CNS" diagnoses group and classifies them as either HGG, LGG, or other, if they don't fall into the glioma category. Note that there are a handful of diagnoses that correspond only to multiplexed samples so ultimately don't get considered, so I ended up just putting `NA` for those diagnoses groups. I also decided that since we are focusing on HGG vs LGG, we don't want to show any non-gliomas in the plot. For those and the normal I labeled as other, but I guess I could just label any samples that won't get included in the plot as NA. 
- I made two functions that I hope will be useful when working on #181 and #182. 
  - The first function just preps the consensus cell types for plotting by calculating the % of each library annotated with a given cell type. It also orders the data frame so that the libraries are ordered by total % of the library that is annotated. Note that I did not end up using `duckplyr` here since I was having issues with some of the functions that I had used when creating this data frame in OpenScPCA and it didn't feel like the time to fight with it just yet. I think since we are not looking at gene expression and just looking at cell type assignments we won't run into the same problems we did with the dot plots. 
  - The second function actually creates the bar plot, with optional faceting. For #181 we won't do any faceting, so it seemed useful to have that option here. 
- Then I added a script that creates the plot specifically for the HGG vs. LGG comparison using those two functions and saves it to a PDF. 

Below is a screenshot of the plot for easy review. I ended up including the library IDs as labels, but I'm not so sure they will remain legible once we get into 

Screenshot 2025-03-13 at 3 25 46 PM

Here's an option without the library IDs. 
Screenshot 2025-03-13 at 3 51 10 PM


sample-info/brain-classifications.tsv
```diff
@@ -0,0 +1,29 @@
+ontology_id	human_readable_value	submitted_diagnosis	diagnosis_group	subdiagnosis_group
+MONDO:0016684	Anaplastic astrocytoma	Anaplastic astrocytoma	Brain and CNS	High-grade glioma
+MONDO:0016700	Anaplastic ependymoma	Anaplastic ependymoma	Brain and CNS	NA
+MONDO:0016734	Anaplastic ganglioglioma	Anaplastic ganglioglioma	Brain and CNS	NA
+MONDO:0021640	Grade III glioma	Anaplastic glioma	Brain and CNS	High-grade glioma
+MONDO:0020560	Atypical teratoid rhabdoid tumor	Atypical teratoid rhabdoid tumor	Brain and CNS	High-grade glioma
```
jaclyn-taroni, line 6: ATRT are embryonal tumors: https://pmc.ncbi.nlm.nih.gov/articles/PMC8328013/#T1

sample-info/brain-classifications.tsv
```diff
@@ -0,0 +1,29 @@
+ontology_id	human_readable_value	submitted_diagnosis	diagnosis_group	subdiagnosis_group
+MONDO:0016684	Anaplastic astrocytoma	Anaplastic astrocytoma	Brain and CNS	High-grade glioma
+MONDO:0016700	Anaplastic ependymoma	Anaplastic ependymoma	Brain and CNS	NA
+MONDO:0016734	Anaplastic ganglioglioma	Anaplastic ganglioglioma	Brain and CNS	NA
+MONDO:0021640	Grade III glioma	Anaplastic glioma	Brain and CNS	High-grade glioma
+MONDO:0020560	Atypical teratoid rhabdoid tumor	Atypical teratoid rhabdoid tumor	Brain and CNS	High-grade glioma
+MONDO:0022965	Desmoplastic infantile ganglioglioma	Desmoplastic ganglioglioma	Brain and CNS	NA
+MONDO:0005499	Brain glioma	Diffuse midline glioma	Brain and CNS	High-grade glioma
+MONDO:0005505	Dysembryoplastic neuroepithelial tumor	Dysembryoplastic neuroepithelial tumor	Brain and CNS	Low-grade glioma
```
jaclyn-taroni, line 9: If a ganglioglioma is not a low-grade glioma, neither is this. Both are glioneuronal and neuronal tumors: https://pmc.ncbi.nlm.nih.gov/articles/PMC8328013/#T1

sample-info/brain-classifications.tsv
```diff
@@ -0,0 +1,29 @@
+ontology_id	human_readable_value	submitted_diagnosis	diagnosis_group	subdiagnosis_group
+MONDO:0016684	Anaplastic astrocytoma	Anaplastic astrocytoma	Brain and CNS	High-grade glioma
+MONDO:0016700	Anaplastic ependymoma	Anaplastic ependymoma	Brain and CNS	NA
+MONDO:0016734	Anaplastic ganglioglioma	Anaplastic ganglioglioma	Brain and CNS	NA
+MONDO:0021640	Grade III glioma	Anaplastic glioma	Brain and CNS	High-grade glioma
+MONDO:0020560	Atypical teratoid rhabdoid tumor	Atypical teratoid rhabdoid tumor	Brain and CNS	High-grade glioma
+MONDO:0022965	Desmoplastic infantile ganglioglioma	Desmoplastic ganglioglioma	Brain and CNS	NA
+MONDO:0005499	Brain glioma	Diffuse midline glioma	Brain and CNS	High-grade glioma
+MONDO:0005505	Dysembryoplastic neuroepithelial tumor	Dysembryoplastic neuroepithelial tumor	Brain and CNS	Low-grade glioma
+MONDO:0019002	Lhermitte-Duclos disease	Dysplastic gangliocytoma	Brain and CNS	Low-grade glioma
```
jaclyn-taroni, line 10: Same comment about ganglioglioma.

jaclyn-taroni:
I need more information about how you did the classifications, @allyhawkins. I would naively expect this table to be our guide: https://pmc.ncbi.nlm.nih.gov/articles/PMC8328013/#T1 I think this is harder to review with using `NA` for multiplexed diagnoses because I can't follow your logic for classification as well.

jaclyn-taroni (commented):
I am going to return my initial comments about the classifications so you can see why I think any `NA` are going to be a problem for review.

allyhawkins:
> I need more information about how you did the classifications, @allyhawkins. I would naively expect this table to be our guide: https://pmc.ncbi.nlm.nih.gov/articles/PMC8328013/#T1 I think this is harder to review with using `NA` for multiplexed diagnoses because I can't follow your logic for classification as well. **Edited to add:** Another approach would be to drop these from the table entirely.

I think I just googled things and totally forgot that this table existed. But I've updated this to hopefully make things more clear and better document how we make these assignments. 

- I added a small script that creates the brain classifications table. This only includes diagnoses for samples that are non-multiplexed that are in the brain and CNS projects. 
- In that script I have a list of HGG vs LGG tumor types where anything in SCPCP000001 is HGG and anything in SCPCP000002 is LGG. All other types are assigned using the table that you sent. I classified any glioneuronal tumors as LGG since ganglioglioma is part of SCPCP000002, which is LGG. ATRT and the focal cortical dysplasia are now labeled as "other" since they are neither HGG nor LGG. 
- I updated the plot based on these changes.

scripts/utils/consensus-celltype-plotting-functions.R
```diff
@@ -189,3 +189,125 @@ marker_gene_dotplot <- function(
...
+    dplyr::summarize(
+      total_cells_per_library = dplyr::n()
+    ) 
+  
+  # get summary stats for each cell type in each library  
+  summary_df <- consensus_df |> 
+    dplyr::left_join(totals_df, by = "library_id") |> 
+    dplyr::group_by(project_id, library_id, sample_id, broad_celltype_group) |> 
+    dplyr::summarize(total_cells_per_annotation = dplyr::n(),
+                     total_cells_per_library = unique(total_cells_per_library),
```
jaclyn-taroni, line 236: The `unique()` is required because the same total number of cells will get repeated per annotation, correct?

jaclyn-taroni (approved):
Looks good! Some minor comments, but I don't necessarily need to see this again.

scripts/utils/consensus-celltype-plotting-functions.R
allyhawkins: Yes that is correct.

scripts/utils/consensus-celltype-plotting-functions.R
```diff
@@ -189,3 +189,125 @@ marker_gene_dotplot <- function(
...
+#' @param facet_variable Column to use for faceting, default is NULL 
+#'
+#' @returns
+#' @export
+#'
+#' @examples
+stacked_barchart <- function(
+    df, 
+    fill_column,
+    celltype_colors, # named vector where names match the values in fill_column
```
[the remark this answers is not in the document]
allyhawkins: Done in https://github.com/AlexsLemonade/scpca-paper-figures/pull/208/commits/255191d0e66db079f3ee3957c307817cf9e854f2

PR #209, opened 2025-03-14
 Add bar chart for HGG vs. LGG highlighting immune celltypes

### LINKED ISSUE #182: Figure 5C: Stacked bar chart for brain/CNS tumors with only immune cells

sjspielman:
Closes #182 
Stacked on #208 

This PR adds code to create a bar plot of hgg/lgg samples with an emphasis on immune cell types. Currently, the code both modifies and builds on code in #208 and will probably experience some organizational changes once this gets reviewed. Specifically, I updated the script to make 5B to make _both_ 5B and 5C due to a decent amount of shared code, and I also added a function specific to preparing data for _this_ plot. There are still some areas where code might be firmed up, but since I am expecting some that there may be some rearrangements anyways upon review (especially if there are decent sized changes in #208 in the first place), I didn't want to change _too_ much more from what's currently there. 

As detailed in #182, I decided not to include granulocytes in the plot as there is only 1 cell in all of the libraries here which has it; this color wouldn't be apparent anyways. In addition, I currently filter to only plot libraries that have at least 20 immune cells, although this threshold can easily be removed or changed.

I added a palette file for this plot's immune cell types as well, and I tried to order the colors so when colors are similar they are less likely to be directly touching in the plot. I used these websites to help me pick colors: https://medialab.github.io/iwanthue/ for initial color selection, and https://mdigi.tools/change-color-hue for assistance refining some of them once I saw how they came out. 

I'm attaching a screen shot of the plot here for convenience! This version does have library_id labels, but I imagine whether those stick around gets sorted out in #208 and I'll apply the result here.
Screenshot 2025-03-14 at 4 21 00 PM


allyhawkins:
Thanks for doing this and doing the color picking! Generally, I think the colors are good, but I had a few questions/comments.

Can you update the legend order so that "other" is last instead of first? 

I also wonder if instead of sorting by total percentage of cells annotated for this plot, we sort by total myeloid percentage? The punchline here is that we see different distributions of myeloid and T cells between HGG and LGG so I wonder if we sort by one of these if that finding will be more clear. 

> In addition, I currently filter to only plot libraries that have at least 20 immune cells, although this threshold can easily be removed or changed.

What does it look like if you don't do any filtering?

sjspielman:
Thanks @allyhawkins, I'll look at these things today!

sjspielman:
@allyhawkins here's how some of these updates would look. What do you think?


### With other on the bottom, and ordering by _fraction myeloid_
Screenshot 2025-03-20 at 10 55 06 AM


### Same filtering as above, and all libraries are shown
Screenshot 2025-03-20 at 10 55 32 AM

allyhawkins:
Sorry, I meant to keep the "other" on the top in the actual plot, but just move where it is in the legend. I realized this is something that needs to be fixed in the `stacked_barchart` function for all the plots, so I'm fixing that as part of https://github.com/AlexsLemonade/scpca-paper-figures/issues/181. 

As for the ordering, I like it! 

I think maybe we should keep the filtering? I feel like it might be misleading to show samples that have 100% myeloid for example, but only a handful of cells that are labeled. That being said, perhaps we should only include samples that have a certain percentage of cells annotated as immune to begin with rather than picking a strict number of cells. So, similar to filtering based on a cell type being present in 1% of cells, maybe we only include samples that have > 1% of cells annotated as immune. Does that make sense?

sjspielman:
> So, similar to filtering based on a cell type being present in 1% of cells, maybe we only include samples that have > 1% of cells annotated as immune. Does that make sense?

This makes sense! I've updated to code to do something like this, but will wait for your next PR to wrap up updates with plot order and re-request review then!

jaclyn-taroni (commented):
This is looking good. I have the same comment as in #218 where we might want to bump up the font size of all but the library identifiers (presumably the best place to do this is in the function making the stacked bar). My assumption is that will help save us time, but as the folks doing the compiling, tell me if I'm wrong.

sjspielman:
> My assumption is that will help save us time, but as the folks doing the compiling, tell me if I'm wrong.

I can't say whether it will really save time or not, but it has the _potential_ to save time, so let's go for it!

> Oh, another idea I have that I don't think I actually like is putting the total number of immune cells somewhere.

The library ids are already long, so on one hand we don't want to make them longer, but on the other hand they're already long so what's a little more text?

sjspielman:
This should now be ready for another look! I made the following changes:

- After merging in changes that were made in #218, plot font is now bigger 👍 
- I made a couple more overall plot sizing changes to both 5B and 5C which should also make compilation easier. 
- But, I also noticed that the changes to how plot scale `breaks` were specified in #218 did alter the order of the legend so it no longer matched the bars. I updated this code to use `levels()` instead of `unique()`, and since it was kind of hard to read (so many parentheses...) I moved that little code bit out of the ggplot code to make and use variable `plot_breaks`. So, pdfs for both 5B and 5C barplots (all cells and immune cells, respectively) were updated with this change made in https://github.com/AlexsLemonade/scpca-paper-figures/pull/209/commits/b0678d29edfd955b6920cbf0672f1d856efbe0f5
- Noting I did not add in number of immune cells, since when I tried it really did make the x-axis labels too long, but we could add a supp table perhaps?

allyhawkins:
I feel like something is off here with the LGG plot? There used to be some space in between the bars so it wasn't just all blocks of colors and the HGG plot looks like it should so I'm not sure what's going on here...  

Screenshot 2025-03-27 at 9 18 02 AM

sjspielman:
> so I'm not sure what's going on here...

This is almost certainly a product of whatever program you are using to view the file and how it's rendering it. Here's a screenshot I see locally:
Screenshot 2025-03-27 at 10 24 33 AM


If you zoom in or out of wherever you are viewing it, it should start looking more reasonable?

allyhawkins:
> > so I'm not sure what's going on here...
> 
> This is almost certainly a product of whatever program you are using to view the file and how it's rendering it. Here's a screenshot I see locally: Screenshot 2025-03-27 at 10 24 33 AM
> 
> If you zoom in or out of wherever you are viewing it, it should start looking more reasonable?

Wow you're right. Opening it with a different PDF viewer shows it correctly, very weird!

scripts/utils/consensus-celltype-plotting-functions.R
```diff
@@ -261,6 +261,127 @@ create_celltype_summary <- function(
...
+  # subset to only immune celltypes, and add column for plotting
+  consensus_df <- consensus_df |>
+    # Keep only the immune cells and remove PDX
+    dplyr::filter(
+      consensus_annotation %in% all_immune_celltypes, 
+      sample_type == "patient tissue"
+    ) |>
+    # Create broad_celltype_group label with value "other" when cells are not T or myeloid types
+    dplyr::mutate(
+      broad_celltype_group = ifelse(
```
allyhawkins, line 317: Very minor, but I don't think we want to call this `broad_celltype_group` because it's different from the `broad_celltype_group` that's being used in 5B.

allyhawkins (commented):
I think the figure looks good, but can you make one change to have the "other" bars on the top of the colored stacked bars rather than on the bottom (the way you had it originally)? That way it matches the other stacked bar plots in 5B and S7.

sjspielman:
I updated that column name as requested, and I rearranged so "other" bars are on the top, but the "other" legend key is on the bottom. When I did this, the "mature alpha-beta T cell"key was right above the "other" legend key, and the colors looked really similar juxtaposed, so I tried to make them a smidge different while still keeping the color scale about what it was. Let me know if you think that worked ok!

Here's the new figure screenshotted:
Screenshot 2025-03-28 at 9 31 15 AM

allyhawkins (approved):
LGTM!

jaclyn-taroni (approved):
LGTM

Issue #183, opened 2025-02-28
Compile consensus cell type figures


allyhawkins:
Once we have created the individual panels for Figure 5 and supplemental figures for consensus cell types, we will need to compile them in illustrator. 

- [ ] Figure 5
- [x] Supplemental Figure 6 - dot plots for all diagnosis groups except brain/cns
- [x] Supplemental Figure 7 - stacked bar charts for all groups except brain/cns

We should do this after we do #178, since that changes the numbers of supplemental figures.

allyhawkins:
Note that when we compile these figures, we should adjust the annotation bar labels for the dot plots so that they are at an angle. We will need to do this in illustrator.

sjspielman:
In https://github.com/AlexsLemonade/scpca-paper-figures/pull/207, we updated the dot plot widths to ensure dots don't overlap, but now panels are different widths. If this proves problematic for compiling, we'll want to reconsider those changes.

PR #210, opened 2025-03-17
Add Figure S6 compilation

sjspielman:
Part of #183 

This PR adds a compiled Figure S6, which is the non-brain/cns dot plots. As part of this, I also updated the code to scale down the maximum size for the dots in the larger panels, since resizing them in Illustrator caused the dots to overlap again.

Currently, the scaling between Panels A and B/C is quite different, but personally I am not bothered by this since this is a supp figure? Panel A is currently exported at a smaller size because it is an overall smaller panel. We could make the export size consistent (e.g. undo some of the changes in #207, which we did note there might be a thing to do!), which would have the effect of adding a ton of whitespace to Panel A. Let me know if we'd like to explore that compilation option as well!


jaclyn-taroni (commented):
I think this is fine for a supplemental figure. Can you update the panel label font to be consistent with everything else we have? I will approve once that's done.

sjspielman:
Labels now in Arial!

jaclyn-taroni (approved):
👍🏻  thanks!

Issue #201, opened 2025-03-12
Expand analysis README


sjspielman:
Now that we have determined which bulk analyses are being used in the manuscript, we should add a bit more detail to `analysis/README` about which analysis are and are not immediately relevant to the manuscript. The goal here is for folks visiting the repository to have a clear understanding about what directly contributed to the manuscript and what didn't.

PR #213, opened 2025-03-17
Get rid of tpm lines

sjspielman:
This tiny PR gets rid of `tpm_dir` creation, which isn't needed anymore and should have been removed in #201!


jashapiro (approved):
⚡

PR #215, opened 2025-03-18
Add standalone sample whitelist and script to generate it

### LINKED ISSUE #214: Create separate script for sample whitelist
This issue is broken out from the second header of this comment: https://github.com/AlexsLemonade/scpca-paper-figures/issues/204#issuecomment-2733499070

We currently create the sample whitelist in each script that this information is needed (Fig 1A, Fig 1B, Table S1). However, the files used to create this sample list are internal metadata files which we will do not plan to release. Therefore, we should add a new script that just generates the sample whitelist file, and that whitelist itself can be released. The scripts that create this whitelist should then be updated to read in the sample whitelist rather than creating it fresh each time. In addition, we will no longer need to read in the (overall scpca) project metadata file in these scripts since it is only used to create the sample whitelist.

sjspielman:
Closes #214 

This PR adds a script to create and save a sample whitelist to the repo in order to ensure others can reproduce our code. There are several changes as part of this:

1. I added a new script in `scripts/figure_setup/prepare-sample-whitelist.R` which creates and exports this list, and I documented in `scripts/README.md`
2. I deleted the former utils script which housed the function to do this; Git decided the changes I made were small enough that it's showing this diff as a rename/modify to create the new script
3. I updated the three scripts which previously created the sample whitelist to now read this in, and no longer source the function to create it. As such, scripts don't need to read in the project metadata anymore either since it was only used to create the whitelist. Note that one of those scripts was reading in the sample metadata which wasnt actually needed, so I removed that line.
4. Finally, I added the whitelist itself to live in `sample-info/sample-whitelist.txt`

Noting at the time of filing this, I am not requesting review since this wasn't necessarily planned work this sprint; I will request review next sprint!


allyhawkins (commented):
I took a look at this and I think the changes look good, just a few small documentation updates. 
Can you update the description in the `scripts/README.md` file to mention that you need to run `prepare-sample-whitelist.txt` prior to running the scripts that you modified here? It should be L61, L64, and L67.

sjspielman:
> Can you update the description in the scripts/README.md file to mention that you need to run prepare-sample-whitelist.txt prior to running the scripts that you modified here? It should be L61, L64, and L67.

While I was beginning to add these comments in, I realized we might want a different approach here altogether for this sort of documentation? The direction we are moving is, nobody _besides us_ should have to run any of the scripts in `figure_setup/` to run any figure scripts. So, including these lines might end up being confusing for reproducibility. Curious to hear your thoughts on that!

allyhawkins:
> > Can you update the description in the scripts/README.md file to mention that you need to run prepare-sample-whitelist.txt prior to running the scripts that you modified here? It should be L61, L64, and L67.
> 
> While I was beginning to add these comments in, I realized we might want a different approach here altogether for this sort of documentation? The direction we are moving is, nobody _besides us_ should have to run any of the scripts in `figure_setup/` to run any figure scripts. So, including these lines might end up being confusing for reproducibility. Curious to hear your thoughts on that!

Ohh yes, that's a good point. I also realize that because we are adding the whitelist to the repo then you wouldn't need to run that script. Maybe we should move the figure setup to be below the list of all the scripts and make a note that these were used to create input files for the script. Not in this PR, but once all of the changes to make metadata/ input data available go in.

allyhawkins (approved):
Approving since we are planning to update the README to account for the figure setup scripts changing at a later time.

sjspielman:
I opened an issue to update the README more generally in #231.

Issue #211, opened 2025-03-17
Add instructions for getting portal manuscript data


sjspielman:
Related to #204 
We would like to write a README file that explains what Portal data is needed to recreate manuscript figures as well as instructions for how the data can be downloaded. This will be included with our upload to Zenodo.

sjspielman:
We will also need instructions for obtaining gene expression: https://github.com/AlexsLemonade/scpca-paper-figures/pull/247#issuecomment-2773718094

PR #217, opened 2025-03-19
Initiate instructions for obtaining and preparing data

sjspielman:
Towards #211 but we may not want to close yet, since there are some TODO items.

The goal of this PR is to initiate the instructions for obtaining and preparing data files. Some aspects are forthcoming, but this lays out the general structure and content. So, I am opening this as a Draft PR since I anticipate building it up further. That said, we could also treat this as a mature PR and make additional changes subsequently, if reviewers have particular thoughts. But a draft seems good to start if only for feedback on the overall presentation and organization!

Broadly, I have two main sections: 
- First, getting data to reproduce figures/tables
  - There are some TODOs in here which still need to be filled out
- Second, getting data to reproduce the analysis
  - I have also included a TSV file which can be used to facilitate organizing files here, since not all samples from the projects are used but if present the code will try to parse them! This TSV gives the samples that are actually included in the analysis and should be retained after downloading data


sjspielman:
I am going to close this in favor of a refreshed PR to address #211, since the landscape has changed a decent amount since this was originally envisioned & drafted.

PR #218, opened 2025-03-20
Supplemental figure 7 barplots for remaining diagnosis groups

### LINKED ISSUE #181: Supplemental Figure with stacked bar charts for all diagnosis groups

allyhawkins:
Closes #181 

Here I'm adding a script to generate the bar plots that show the consensus cell type assignments for all other diagnosis groups besides Brain and CNS. The plots being added here will be compiled to create supplemental Figure 7. 

- Because there are a lot of samples in each group, mostly in the Leukemia group, I wonder if we want to either not show the library IDs or break these up by diagnosis (or maybe project)? I think showing all the samples in one plot is going to make the library IDs impossible to read when we compile everything. 
- While working on this, I made a small change to the legend for these plots to always show "unknown" or whatever the grey color will be at the end. I added an argument to the `stacked_barchart` function to account for this so that we can use it with  the "other" group in the immune bar plots being added in #209. Because of that I had to regenerate the HGG vs. LGG plot.


allyhawkins:
@jaclyn-taroni I've been playing around with this and want to get your opinion before I start reorganizing code more. Because each of the diagnoses have a different number of samples that varies from just a handful to > 50, just faceting by diagnosis looked bad. Either the library IDs were too squished with multiple columns for diagnoses with a large number of samples or using just one column meant the smaller groups were really spread out. So I came up with a way to dynamically assign the number of columns in each row of the facet.

We also have a lot of diagnoses that only have a handful of samples, so I decided to take anything with less than 10 samples and group into "other". Each panel of the figure would still be specific to a diagnosis group so there would be one "other" for each diagnosis group. 

Here's an example of what this looks like for each of the groups. 

### Leukemia 

Screenshot 2025-03-21 at 12 53 11 PM

### Sarcoma 

Screenshot 2025-03-21 at 12 54 52 PM

### Other solid tumors 

Screenshot 2025-03-21 at 12 59 33 PM

Let me know what you like/ don't like and I'll update the code accordingly.

jaclyn-taroni:
These LGTM!

allyhawkins:
Alright @jaclyn-taroni the code has now been updated and the plots re-rendered to use the faceting I mentioned in my previous comment. I ended up mostly keeping the original function for creating the stacked plots so that we can still use that for 5B and 5C and then wrote a new function that does the dynamic faceting and combines plots for each diagnosis group. I also was able to adjust the font size in the main function for creating the stacked bar plots. 

This should be ready for another look.

jaclyn-taroni (approved):
LGTM 👍🏻

PR #222, opened 2025-03-21
Use SCPCL000498 for both panels in S5

### LINKED ISSUE #220: Use the same sample for both panels in S5
I noticed that in Supplemental figure 5 we are using example plots from the cell type report for two different samples with no real reason to do that. I must have missed that they were different when condensing the figures so we should update this to include the same sample. Looking at the reports, I think it would make sense to use `SCPCL000498` for both plots.

allyhawkins:
Closes #220 

Here I updated panel A in S5 to use SCPCL000498, which is the same library that we use for panel B. All I had to do was change the sample/library ID in the script and then re-compile the figure.


jaclyn-taroni (approved):
👍🏻

PR #226, opened 2025-03-26
Compile S7 and update fonts in S6

allyhawkins:
Towards #183 

Here I am compiling all the stacked bar charts for non brain and CNS tumors. I didn't really do anything special other than making sure everything was arial font (to match the rest of the figures) and then I made sure sizes matched where applicable. The only place the sizes doesn't match is in the B-ALL panel since there are so many samples. I used the biggest size that was legible. 

While working on this I was looking at S6 and noticed that the font didn't match the rest of the figures and that we had wanted to angle the labels for the annotation bars at the top. So I made those adjustments here. 
Edit: I also added some panel titles with the diagnosis group.


jaclyn-taroni (approved):
👍🏻

PR #232, opened 2025-03-27
Add compiled bulk figures: 6 and S8

### LINKED ISSUE #227: Create compiled version of bulk figures
It looks like we have the bulk figure panels output in `figures/pdfs`, but we do not have final compiled version of the figures. Mainly, we need to add a copy of the compiled version to `figures/compiled_figures` and add panel labels (A. and B.). We should also make sure that the fonts are all Arial to match the rest of the figures while we're in illustrator.

sjspielman:
Closes #227 

This PR adds compiled versions of Figure 6 and S8. The fonts seemed like they were already good here, so for this compilation I just ensured size was within 8.5x11 and added panel labels A/B.


jaclyn-taroni (commented):
You need periods after the panel letters to match the rest of the figures

sjspielman:
Woops! Periods added!

jaclyn-taroni (approved):
LGTM 👍🏻

PR #225, opened 2025-03-21
Export S1B-D benchmarking data TSVs

sjspielman:
Closes #216 

This PR adds code to the `FigS1B-D_method-metrics-comparison.R` script to export TSV files needed to remake figures to assist in reproducibility. TSVs are exported to a new directory `reproduce-figures-analysis`, which is also being populated in other PRs (#217, #224 for now) - because this directory is subhect to change and I want to reduce conflicts here, I haven't yet documented these files. I do plan to open a future issue to solidify this directory later though.

**Notably** it was quick to add an `if` statement to control which aprts of the script to run, so I went ahead and did it. But, this was all in one commit (https://github.com/AlexsLemonade/scpca-paper-figures/commit/cc4fa9da11c6c7704a9db4daec24b29315582da5) and can be reverted for just a simple TSV export.

PR #224, opened 2025-03-21
Add metadata script

sjspielman:
Stacked on #217 
Closes #212 

This PR adds a script to convert the portal-wide sample metadata tsv to sample and library metadata files which can be used to reproduce code. This is being opened as a Draft, because we would like to test the script before merging it in. Once it is able to be fully tested and is successfully run, we can merge this (likely closer to submission).

PR #237, opened 2025-03-31
Review first: Establish reproducibility directory

### LINKED ISSUE #236: Establish directory to hold data and instructions for reproducibility
I have a few PRs that are sending files into a directory meant to stash:

- Instructions for obtaining and preparing data for reproducibility (#217)
- A script for preparing metadata that can be used for reproducibility (#224)
- Datasets necessary for reproducibility (#225)

This is perhaps not an ideal order to go about things! A PR to establish said directory would be ideal to review first. This issue tracks making that directory.

sjspielman:
Closes #236 

After filing several prep files to live a in directory to facilitate reproducibility, I realized (better late than never!) that we'd really be better served making that directory first! 

This is a pretty small PR that does that and adds a small readme that will be built up with:
- #217 
- #224 
- #225

### DIFF
--- a/reproduce-figures-analysis/README.md
+++ b/reproduce-figures-analysis/README.md
@@ -1,0 +1,1 @@
+This directory contains instructions as well as necessary scripts and data for reproducing manuscript figures, tables, and analyses.
--- a/reproduce-figures-analysis/README.md
+++ b/reproduce-figures-analysis/README.md
@@ -1,1 +1,0 @@
-This directory contains instructions as well as necessary scripts and data for reproducing manuscript figures, tables, and analyses.
--- a/reproduce-figures/README.md
+++ b/reproduce-figures/README.md
@@ -1,0 +1,1 @@
+This directory contains instructions as well as necessary scripts and certain data files for reproducing manuscript figures and tables.


reproduce-figures-analysis/README.md
allyhawkins, line 1: What about `reproduce-figures` since there is no analysis actually being done here?

sjspielman: Yeah, I agree this is wordy. The reason I have `analysis` in there is because there are also instructions in #217 which discuss how to prepare files in the `analysis/pseudobulk-bulk-prediction` directory to reproduce those too. 

I think we could split up these instructions so instructions for analysis reproducibility are separate from this directory, and instead we can document that in the analysis README itself. This would be fine for reproducing figures since all the necessary files from the bulk analysis for this are either in the repo already or will be added to Zenodo (see #244). What do you think of that organizational approach?

allyhawkins: > I think we could split up these instructions so instructions for analysis reproducibility are separate from this directory, and instead we can document that in the analysis README itself. This would be fine for reproducing figures since all the necessary files from the bulk analysis for this are either in the repo already or will be added to Zenodo (see https://github.com/AlexsLemonade/scpca-paper-figures/issues/244). What do you think of that organizational approach?

This sounds good to me.

allyhawkins (approved):
🚀

PR #238, opened 2025-03-31
Add compiled figure 5

### LINKED ISSUE #183: Compile consensus cell type figures

sjspielman:
Closes #183 

This PR adds the compilation of Figure 5.
To achieve this, I also did a minor code update to allow for a more flexible y-axis label for the barplot so I could specify that the 5C y-axis was percent of _immune_ cells.

When compiling this, I ensured all fonts were Arial, and I was able to bump some dot plot font sizes. They still won't be as big as the barplot fonts, but any little bit counts! Let me know what additional updates I can make here.


allyhawkins:
Thanks for doing this! Can you update the labels for the annotation bar to be angled? I used 330 for the angle in S6.

jaclyn-taroni (approved):
LGTM 👍🏻

PR #246, opened 2025-04-01
Add metadata script

### LINKED ISSUE #212: Add script to create individual metadata files from large metadata download
Most of our scripts use one of our internal metadata files, `scpca-{project,library,sample}-metadata.tsv` as part of figure or table generation. We do not plan to release these internal metadata files, but we do want to provide others with an approach to reproduce our figures nonetheless.

To this end, we plan to write a script that can parse and convert the Portal-wide metadata download tsv into versions of these three metadata files that can be used to reproduce figures.

sjspielman:
Closes #212 

This PR adds a script that converts the portal-wide metadata into tsvs that can be used with our figures and analysis code. I added some documentation for it in the README as well, which references the full instruction file coming up in #217.

### DIFF
--- a/reproduce-figures/README.md
+++ b/reproduce-figures/README.md
@@ -1,1 +1,4 @@
 This directory contains instructions as well as necessary scripts and certain data files for reproducing manuscript figures and tables.
+
+* `prepare-metadata-files.R` is a script which can be used to generate metadata files needed to reproduce figures and tables, as described in `obtain-prepare-data.md`.
+*
--- a/reproduce-figures/README.md
+++ b/reproduce-figures/README.md
@@ -1,4 +1,3 @@
 This directory contains instructions as well as necessary scripts and certain data files for reproducing manuscript figures and tables.
 
 * `prepare-metadata-files.R` is a script which can be used to generate metadata files needed to reproduce figures and tables, as described in `obtain-prepare-data.md`.
-*
--- a/reproduce-figures/prepare-metadata-files.R
+++ b/reproduce-figures/prepare-metadata-files.R
@@ -1,0 +1,58 @@
+# This script can be used to create versions of `scpca-sample-metadata.tsv` and `scpca-library-metadata.tsv` needed to
+# reproduce manuscript figures, tables, and analyses.
+#
+# To run this script, you will first need to download all sample data from the ScPCA Portal: <https://scpca.alexslemonade.org/>
+# From this page, click the `Get All Sample Metadata` button on the top-right of the page.
+# Provide this TSV file as input to this script as:
+#
+# Rscript prepare-metadata-files.R --portal_sample_metadata <path to metadata.tsv>
+#
+# Files will be exported to the directory the script is run from, but this can be specified with `--output_dir`:
+#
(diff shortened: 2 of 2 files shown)


reproduce-figures/README.md
```diff
@@ -1 +1,3 @@
 This directory contains instructions as well as necessary scripts and certain data files for reproducing manuscript figures and tables.
+
+* `prepare-metadata-files.R` is a script which can be used to generate metadata files needed to reproduce figures and tables, as described in `obtain-prepare-data.md`.
```
allyhawkins, line 3: I'm assuming this file is being added in the instructions PR?

reproduce-figures/prepare-metadata-files.R
```diff
@@ -0,0 +1,58 @@
...
+option_list <- list(
+  make_option(
+    "--portal_sample_metadata",
+    type = "character",
+    help = "Path to TSV of all sample metadata downloaded from ScPCA Portal"
+  ),
+  make_option(
+    "--output_dir",
+    type = "character",
+    default = ".",
```
allyhawkins, line 27: I think we might actually want to make a default folder for this to keep things organized. Maybe `modified-portal-metadata`?

reproduce-figures/prepare-metadata-files.R
```diff
@@ -0,0 +1,58 @@
...
+# Define output file paths
+sample_metadata_tsv <- file.path(opts$output_dir, "scpca-sample-metadata.tsv")
+library_metadata_tsv <- file.path(opts$output_dir, "scpca-library-metadata.tsv")
+
+# Read input metadata file
+portal_metadata <- readr::read_tsv(opts$portal_sample_metadata)
+
+# Create sample metadata table
+sample_metadata <- portal_metadata |>
+  dplyr::select(scpca_project_id, scpca_sample_id, diagnosis, disease_timing, is_cell_line) |>
```
allyhawkins, line 48: I'm assuming these are the only columns we use, but I think we'll find out once you run this through to regenerate figures!

allyhawkins (approved):
I think this looks good and I think we'll know more about whether or not the columns we have are sufficient when you try and reproduce all the figures, so I'm going to approve. 

My only comment about this is saving the output to a directory by default rather than the directory that the scripts live in? I feel like that would help just keep things organized for someone running it, but I'll leave the final decision up to you.

reproduce-figures/README.md
sjspielman: Yes, that's the main component to that PR!

reproduce-figures/prepare-metadata-files.R
sjspielman: Yes, this is what I found when looking across scripts that only these columns were needed. We'll find out soon enough if some are missing!

reproduce-figures/prepare-metadata-files.R
sjspielman: What about just `metadata-files`? I suggest that specifically because it matches the script name.

allyhawkins (approved):
LGTM

PR #247, opened 2025-04-02
Add script to prepare consensus cell types for Zenodo

### LINKED ISSUE #240: Prepare consensus cell types to release
Broken out of #204 

We have decided to disseminate the consensus cell type results as they are formatted & exported from `OpenScPCA-nf`. We may also want to try and reduce size by slicing out columns. 
This issue tracks preparing this dataset and associated docs for dissemination. Notably, we expect this to be too big to be stored in the repo, so we'd to make this a separate Zenodo upload. Therefore, as part of review here, I will upload the data intended for Zenodo to our Google Drive.

sjspielman:
Closes #240 

This PR adds a script which prepares consensus cell types for eventual upload to Zenodo. I placed this script in `figure_scripts/` with comments at the top describing usage. 
This script will read in all consensus cell type files and create a compressed (gz) directory with a subsetted version of those files, considering only columns used in the code. I determined which columns to use based on code in these scripts.

- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/scripts/Fig5B-C_brain-barplots.R (uses util functions above)
- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/scripts/FigS7_consensus-bar-plots.R
- util functions used by the above 2 scripts: https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/scripts/utils/consensus-celltype-plotting-functions.R
- https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/analysis/pseudobulk-bulk-prediction/scripts/prepare-ora-gene-sets.R

Notably these columns don't appear to be directly used but seemed worth keeping so I kept them: `sample_id`, `project_id`, and `barcodes`.

I have uploaded the output to Google Drive for review: https://drive.google.com/drive/u/0/folders/13IfW8WUWQol8aUASzVqNiJ1fzinCQHiQ
That said, it's worth noting that the file is 27 Mb, which is small enough to live in the repo if we want it to!


allyhawkins (commented):
This looks good, other than one small comment about the error checks. 

It took me a second to figure out why the consensus cell type files that we copy over were so much bigger than what you are seeing here and that's because we also have the gene expression files that are used to make the dotplots... Technically all of that data is available on the Portal, so we don't really need to provide it and the marker gene lists are all directly pulled from `OpenScPCA-analysis`. It is a pain for someone to have to download everything from the Portal and then grab the expression for specific marker genes to be able to make the dot plots, but it's not impossible. I think if its something that's allowed with our T&C we might also want to save the gene expression TSVs alongside the consensus cell types?

sjspielman:
>  that's because we also have the gene expression files that are used to make the dotplots.

This is definitely my bad then - I didn't realize those files were being used so didn't check for them, but it clearly makes sense that they are used for dot plots! I will update this code to include both files for sure, and once that happens we're _definitely_ not keeping this in repo!

sjspielman:
This PR has been roundly overhauled! Since we in fact need both consensus files but only the TSVs are modified, I figured it would be more straightforward as a bash script. I removed the R script and added a bash script that takes the following steps:

- remove any existing output files if they exist
- copy the entire consensus cell type results directory
- in the new copy, loop over the TSVs and cut out only columns of interest and re-save
- compress the whole darn thing ---> 4.4 GB. This is currently uploading to the google drive.

sjspielman:
> I think if its something that's allowed with our T&C we might also want to save the gene expression TSVs alongside the consensus cell types?

That said, before we definitely proceed with this bash file vs revert to the Rscript, let's confirm about T&C when @jaclyn-taroni has a chance to weigh in.

jaclyn-taroni:
I would not want to distribute gene expression data via Zenodo.

allyhawkins:
> I would not want to distribute gene expression data via Zenodo.

@sjspielman Probably best to revert the changes to export the expression TSV and instead make a note about mentioning getting gene expression in the instructions PR you have open?

sjspielman:
I'm closing out this PR to pursue an updated plan laid out in #250

PR #249, opened 2025-04-02
Filter to only samples used in analysis

### LINKED ISSUE #241: Update bulk analysis code with a library whitelist
Currently, the code in `pseudobulk-bulk-analysis` relies on presence/absence of synced ScPCA data files for analysis. This poses challenges for reproducibility by others, since code will map over all data files regardless of whether those data files are meant to be used in analysis. When someone outside the Data Lab attempts to run the code, we want them to be able to download SCE files from the Portal for all projects without worrying that some are not used. 

The best way to approach this is probably to include a whitelist of libraries, at each modality, which should be part of analysis and ensuring the code uses this whitelist to not consider libraries we don't use. We already have a file mapping bulk library ids which are used, so we might either expand that file or make a separate one to also house single-cell library ids, and then update code accordingly.

sjspielman:
Closes #241 

The overall goal of these changes is to update code to _not assume_ that all libraries present should be used in analysis. This is because, when others download portal data to reproduce, they should be able to keep all the samples from the portal without having to worry which are actually used in the bulk analysis.

Implementing this turned out to be simpler than I expected! We already have a tsv file with bulk libraries & samples that are used in analysis, so we can use those sample ids to filter down to single-cell/nuclei samples of interest too. The only place this needed to happen was in the script that calculates pseudobulk. After this stage, the scpca data files are no longer used in the code. I added this map file as an input argument to the script and use it to subset samples before doing pseudobulk calculations.

I tested this out by downloading `SCPCP000009` (full of unused multiplexed!) from the portal and ran the script with that as the data input directory. When I ran the code, the resulting pseudobulk tsv only contained the 4 (non-multiplexed) samples we use for this analysis, so it's working as expected!

### DIFF
--- a/analysis/pseudobulk-bulk-prediction/run-prediction.sh
+++ b/analysis/pseudobulk-bulk-prediction/run-prediction.sh
@@ -43,6 +43,7 @@ mkdir -p $ora_html_dir
 consensus_celltype_dir="../../s3_files/cell-type-consensus-results"
 
 # This convenience file keeps track of the bulk library & sample ids used
+# This is also used to assist filtering to single-cell processed RDS files of interest based on sample
 map_file="${data_dir}/bulk-library-sample-ids.tsv"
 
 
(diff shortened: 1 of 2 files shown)


allyhawkins (approved):
🐈

PR #251, opened 2025-04-07
Add script to prepare all the scpca data

### LINKED ISSUE #250: Add script to parse portal data
Broken out of #240 based on this discussion: https://github.com/AlexsLemonade/scpca-paper-figures/issues/240#issuecomment-2775625842

Given that others will indeed need to download the whole portal, this issue tracks writing a script to prepare ScPCA data files for reproducibility. Initially, the primary goal of this script was to prepare files in `s3_files/cell-type-consensus-results/`, but considering all items they'll need from the portal, it does seem worth it for this script to arrange  _all_ the ScPCA Portal data needed for reproducing both analysis and figures. This seems worthwhile because the additional steps beyond parsing out consensus cell type TSVs are just copy/pasting files around, which should not be a big lift to code up. 

This script would therefore take as input:
- The path to a directory containing all the downloaded portal project directories (unmerged rds files!), ~named by project id only (this means they will have to do some directory renaming on their own, which seems fine).~ **edit:** they don't need to rename! less manual more better 🎉 
- The path to a separate download, the merged SCPCP000003 object

The script should output two directories by default, which users can place in the appropriate location.
- `s3_files/` with scpca portal data arranged as expected by scripts
- `scpca_data/`, which is the directory in `analysis/pseudobulk-bulk-analysis/data/` that contains portal data needed for analysis - the processed SCE files & bulk count TSVs for the 5 projects analyzed

This script will also require a decent amount of documentation. I expect this PR to contain docs within the script itself, and #211 next will provide more comprehensive instructions in context of additional steps needed to prepare data for reproducibility.

sjspielman:
Closes #250 

The goal of this PR is to provide a script that can prepare _all_ the data derived from the ScPCA portal needed for reproducibility. This is being opened as a draft since consensus cell types aren't actually in the portal yet (https://github.com/AlexsLemonade/scpca-nf/issues/837). But, I coded this up _for now_ in a way that it should run and it can be reviewed. 

Here are the PR changes, and please let me know where I can clarify anything here!!

### main changes

- I added a script `reproduce-figures/prepare-scpca-portal-data.R` which creates an output directory containing two directories: `s3_files` and `scpca-data`. The former will contain ScPCA data for the figures, including all the SCEs we use and consensus cell type files, and the latter will contain ScPCA data for bulk analysis including processed SCEs and bulk quant tsvs.
- There are some docs at the top of this script, which I expect to be expanded when we finally make our way to complete #211. Note that, along these lines, I also updated docs to say throughout that reproducibility instructions are in the README file. Since instructions will now be much shorter now that this script takes care of most moving parts, it made more sense to me to just keep all the info they need to read in README - after all it's called "read me" for a reason! :) 
- The script proceeds as follows:
  - First, check that all project zip files are present in a provided input directory, and that the merged SCE is available at a provided path
  - Copy the merged SCE into the `s3_files` output directory
  - Then, map over the projects to, for each sample directory, copy RDS files where they need to go and remove extraneous ones
  - Then, map over the projects to, for each sample directory, do all the parsing for consensus cell types. There is a _major TODO in this code_: consensus cell types aren't actually on the portal yet, so FOR NOW I am using the singler annotation column instead of consensus. Before merging, we will need to replace that column name. Hence, I am opening this as a Draft, although the code can be reviewed now.
- One aspect of the code that I'm pretty sure is correct but would love another set of eyes to confirm, is that we don't need to be keeping around any multiplexed samples since the figures that use ScPCA portal data don't consider multiplexed as far as I can see. Am I missing a spot or is this indeed the case?

### incidental additional changes

- The path we had for the merged SCE doesn't exist anymore, so I updated it in `scripts/figure_setup/sync-data-files.R`
- I added a shebang to `reproduce-figures/prepare-metadata-files.R`

### testing the code

I have tested the code by downloading a few projects from the portal and running through; note there is a check that I had to manually bypass that checks if all 23 projects (aka those given in the project whitelist) are actually present. Therefore, you too can bypass this check and test the code with just one project if you like! You can download a project from the portal, and maybe also the merged SCPCP000003 SCE, and follow the directions I have at the top of the script to run a mini version of it. This is also a review of those directions themselves.


sjspielman:
Noting https://github.com/AlexsLemonade/scpca-paper-figures/issues/253 has been opened to ensure we get the correct column name in here in the end! I'm therefore going to mark this ready for review.

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+#     --portal_projects_dir <path to directory with all zip files> \
+#     --merged_sce_path <path to merged SCE file> \
+#     --output_dir <output directory to save prepared files>
+# 
+# If `--output_dir` is not provided, it will default to `./scpca-portal-data`.
+# Note that this script will not overwrite an existing output directory without the `--overwrite` flag.
+#
+# Finally, move the directories created in the provided output directory to their final locations:
+# - `s3_files` should be placed in the top-level of the `scpca-paper-figures` repository
+# - `scpca-data` should placed inside the bulk analysis data directory, `scpca-paper-figures/pseudobulk-bulk-prediction/data/`
```
allyhawkins, line 35: I think we want to minimize any moving of folders that someone would have to do. So perhaps we should set the default directory to be the root of the repo?

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+  ),
+  make_option(
+    "--merged_sce_path",
+    type = "character",
+    help = "Path to the merged SCPCP000003 project file as a zip file"
+  ),
+  make_option(
+    "--output_dir",
+    type = "character",
+    default = "./scpca-portal-data",
```
allyhawkins, line 61:
```suggestion
    default = here::here(), 
```
I think this should work?

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+
+# Define input files ------------------------
+
+# Read whitelist so we can check that all projects are present
+expected_projects <- readLines(opts$project_whitelist)
+
+# Define and check input projects
+input_zips <- list.files(
+  opts$portal_projects_dir, 
+  pattern = "^SCPCP\\d{6}_SINGLE-CELL_SINGLE-CELL-EXPERIMENT_\\d+{4}-\\d+{2}-\\d+{2}\\.zip",
```
allyhawkins, line 156: I'm wondering if it makes sense to update this so it just looks for the project id and then single-cell? At some point in the near future we are updating the download structure from the Portal (see https://github.com/AlexsLemonade/scpca-docs/issues/383 or https://github.com/AlexsLemonade/scpca-docs/issues/381), so I wonder if there is a way to make this script robust to those changes while also working with the current portal?

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+
+
+# Map over project zip files to organize files for reproducibility ------------------
+input_zips |>
+  purrr::iwalk(
+    \(project_zip, project_id){
+      
+      # temporary variables used for testing
+      #project_id <- "SCPCP000001"
+      #project_zip <- "/Users/sjspielman/Downloads/projects/SCPCP000001_SINGLE-CELL_SINGLE-CELL-EXPERIMENT_2025-04-03.zip"
```
allyhawkins, line 185: Reminder to remove these lines

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+              system(glue::glue("cp {sample_dir}/*processed.rds {output_dir_celltype_results}"))
+            } 
+            
+            # Copy rds files to standalone sample dir
+            if (sample_id %in% standalone_sample_dirs) {
+              system(glue::glue("cp -r {sample_dir} {output_dir_s3_files}"))
+            } 
+          
+            # We can now remove the (un)filtered files
+            system(glue::glue("rm {sample_dir}/*filtered.rds"))
```
allyhawkins, line 223: I think if we are going to remove files then it probably makes sense to just remove the unzipped folder after doing all the copying rather than removing individual files?

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+        purrr::walk(
+          \(sample_dir) {
+              
+            # Define input and output files
+            processed_rds_file <- list.files(
+              path = sample_dir,
+              pattern = "_processed\\.rds$", 
+              full.names = TRUE
+            )
+            stopifnot("Expected a single processed rds file." = length(processed_rds_file) == 1)
```
allyhawkins, line 245: What about when there are multiple libraries per sample? Then we have two processed files.

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+            
+            # Read in the processed SCE
+            sce <- readRDS(processed_rds_file)
+            
+            # Prepare and export the consensus cell types table
+            data.frame(
+              project_id = metadata(sce)$project_id, 
+              sample_id  = metadata(sce)$sample_id, 
+              library_id = metadata(sce)$library_id, 
+              sample_type = metadata(sce)$sample_type,
```
allyhawkins, line 264:
```suggestion
              sample_type = paste0(metadata(sce)$sample_type, collapse = ",")
```
This is going to cause you issues as is for the samples that have multiple sample types.

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,309 @@
...
+    \(project_zip, project_id){
+      
+      # temporary variables used for testing
+      #project_id <- "SCPCP000001"
+      #project_zip <- "/Users/sjspielman/Downloads/projects/SCPCP000001_SINGLE-CELL_SINGLE-CELL-EXPERIMENT_2025-04-03.zip"
+      
+      project_consensus_dir <- file.path(output_dir_consensus_files, project_id)
+      
+      # Unzip the directory, specifying the target in consensus
+      unzip(project_zip, exdir = project_consensus_dir)
```
allyhawkins, line 190: Okay after looking at this whole script, what I think I might do is create a scratch directory to hold the unzipped project, then copy over any processed SCE files/ files needed for bulk, and then read in the SCE file and output the TSVs to the consensus project directory. Then at the end you remove the scratch directory. I think that should simplify some of the code.

scripts/figure_setup/sync-data-files.R
```diff
@@ -65,7 +65,7 @@ sync_call <- glue::glue("aws s3 sync '{celltype_s3_files}' '{celltype_local_file
 system(sync_call)
 
 # sync merged object results for SCPCP000003 -----------------------------------
-merged_s3_dir <- "s3://nextflow-ccdl-results/scpca/processed/results/merged/SCPCP000003"
+merged_s3_dir <- "s3://nextflow-ccdl-results/scpca-prod/results/SCPCP000003/merged/"
```
allyhawkins, line 68: Let's not change this. This is a really big object and I was using the smaller one that had the specific libraries that we need to make this figure.

allyhawkins (commented):
This mostly looks good, I just have a few comments that might help clean things up.

reproduce-figures/prepare-scpca-portal-data.R
sjspielman: Yeah, that's reasonable. We can set things up with `here`!

reproduce-figures/prepare-scpca-portal-data.R
sjspielman: I see I should have simply scrolled down instead of writing this comment. https://github.com/AlexsLemonade/scpca-paper-figures/pull/251#discussion_r2033558243

reproduce-figures/prepare-scpca-portal-data.R
sjspielman: Great point! On it. We'll still have to exclude merged, but we do expect that string to stay in the file name ✔️ . Not clear to me from those issues if we'll still have names in caps or switch to lowercase, so will make this code flexible for either.

scripts/figure_setup/sync-data-files.R
sjspielman: I definitely get that, but `s3://nextflow-ccdl-results/scpca/processed/results/merged/SCPCP000003` doesn't exist anymore so this line doesn't do anything, and it won't copy down a smaller object.

```
(base) ~ ➤ op run -- aws s3 ls s3://nextflow-ccdl-results/scpca/processed/results/merged/SCPCP000003/                       
(base) ~ ➤                                                                                                   
```

reproduce-figures/prepare-scpca-portal-data.R
sjspielman: Looking through our library metadata right now, and I'm not sure what circumstance this would occur in _besides_ multiplexed, which again as I am understanding things, we don't need data for and has already been removed (am I correct though?). 

When I look at duplicate library ids, it's always 10x & cite, or 10x & cellhash, and 10x & cite is still only 1 processed file. So, I'm not sure what scenario we'd expect more than 1 here. What am I missing?

allyhawkins: > When I look at duplicate library ids, it's always 10x & cite, or 10x & cellhash, and 10x & cite is still only 1 processed file. So, I'm not sure what scenario we'd expect more than 1 here. What am I missing?

I'm not talking about duplicate library IDs, I'm talking about the scenario where we have multiple unique libraries per sample. There's not always a 1:1 ratio for sample to library. Here you are getting all `_processed.rds` files in a sample directory and if we have multiple libraries with the same sample ID then you will get multiple `_processed.rds` files that each will need to be read in to grab consensus cell types.

sjspielman:
This has been roundly updated!! Here's a summary of the changes:

- I now use a `scratch` directory to unzip into, and copy files out from there
  - I do still have some lines of code to remove the (un)filtered (and html, because i was removing things anyways) when copying over to bulk, because we want to be cognizant of disk space here. These files aren't used by the bulk analysis, and they aren't small either, so it seemed best to not force people to have them.
  - The `scratch` directory is ignored, except for a `.gitkeep`. I want to note though, we could use a temp dir for this instead of literally scratch. I appreciated using scratch when writing this code because seeing what was going on clearly helped me write the code, but now that it works (ps: can confirm, it works!), we could replace with a tempdir and remove scratch. What do you think?
- By default, the output directories are the directories that they are "supposed" to be! But, I do include arguments to customize them if someone really wanted/needed...eg me when developing the code 😉  
- I created a new `utils.R` file with functions for the consensus cell type processing, which helps make the script more manageable since it's already on the long side and we're expecting a few more steps in #254. 
- Docs have generally been updated in this script and in the README, but again this README will see many more updates in #211.
  - I tried to emphasize caution here! I estimated space because the full set of zipped downloads plus the largest unzipped project is about 164 GB (one project at a time will be unzipped in scratch, so that's the worst case size scenario as it were), so 170 seemed like a good recommendation.

.gitignore
```diff
@@ -10,6 +10,9 @@ s3_files
 #DS Store
 .DS_Store
 
-# Illustrator files 
+# Illustrator files
 *.ai
 
+# scratch
+reproduce-figures/scratch/*
+!reproduce-figures/scratch/.gitkeep
```
allyhawkins, line 18:
```suggestion
!reproduce-figures/scratch/.gitkeep

```

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -0,0 +1,299 @@
...
+    "--s3_files_dir",
+    type = "character",
+    default = here::here("s3_files"), 
+    help = "Output directory for `s3_files`. Default is in the top-level of the repository, which is where code expects it to be."
+  ),
+  make_option(
+    "--bulk_data_dir",
+    type = "character",
+    default = here::here("scpca-data"), #here::here("analysis", "pseudobulk-bulk-prediction", "data", "scpca-data"), 
+    help = "Output directory for bulk data. Default is `analysis/pseudobulk-bulk-prediction/data/scpca-data`, which is where code expects it to be."
```
allyhawkins, line 73: You have one default in `default` and mention a different one in `help`?

reproduce-figures/utils.R
```diff
@@ -0,0 +1,69 @@
...
+
+
+
+#' Prepare and export TSV of consensus cell types
+#'
+#' @param metadata Metadata slot of an SCE object 
+#' @param sce_coldata Coldata slot of an SCE object 
+#' @param output_tsv Path to output TSV file
+#'
+prepare_consensus_tsv <- function(sce_metadata, sce_coldata, output_tsv){
```
allyhawkins, line 13:
```suggestion
prepare_consensus_tsv <- function(sce, output_tsv){
```
I think it would be more straight forward to just read in the sce and grab the col and metadata in the function.

reproduce-figures/utils.R
```diff
@@ -0,0 +1,69 @@
...
+      use.dimred = FALSE
+    ) |>
+      tidyr::pivot_longer(starts_with("ENSG"), names_to = "ensembl_gene_id", values_to = "logcounts") |>
+      # add library id column
+      dplyr::mutate(library_id = metadata(sce)$library_id, .before = 0)
+  }
+  
+  # export the marker gene expression tsv
+  readr::write_tsv(gene_exp_df, output_tsv)
+}
```
allyhawkins, line 69:
```suggestion
}

```

allyhawkins (commented):
I like the changes that you made and think keeping it as `scratch` is totally fine but ultimately up to you if you want to change it to a temp directory. Either way it gets removed so it's just to hold files while unzipping. 

I had a few minor comments that I think should get addressed before merging.

reproduce-figures/prepare-scpca-portal-data.R
sjspielman: this is some kind of copy/paste woopsie...

sjspielman:
I'm perfectly happy with scratch myself, so let's keep it. Got the other reviews in!

allyhawkins (approved):
LGTM 🚀

PR #256, opened 2025-04-09
Consolidate into a single reproducibility script

### LINKED ISSUE #254: Consolidate reproducibility scripts
Following #251, we will have two scripts in `reproduce-figures` and one additional script/step will be needed:

- one to prepare/parse portal SCE files
- one to prepare library & sample metadata files
- in addition to these items, we'll also need to document how to obtain files currently in `s3_files/references`, which are a few files from our public `s3://scpca-references` bucket.

It would be worth popping this all in one script to make this process as smooth as possible. This will entail modifying the script in #251 in two ways:

- Add steps to create the metadata tsvs. This code will be taken from the existing script (which will be removed) to parse metadata, and it will require one additional script argument: the path to the downloaded portal metadata tsv
- Add steps to sync S3 files. We'll need to `system()` the following lines, but the target directory should be `s3_files/references`.
  ```
  aws s3 cp s3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.spliced_cdna.tx2gene.tsv . --no-sign-request
  aws s3 cp s3://scpca-references/homo_sapiens/ensembl-104/annotation/Homo_sapiens.GRCh38.104.mitogenes.txt . --no-sign-request
  aws s3 cp s3://scpca-references/celltype/singler_models/ . --recursive --exclude "*" --include "*.rds" --no-sign-request
  ```

Then, we can proceed to test it all out and write up instructions for using this _single_ script to prepare all the input data needed.

sjspielman:
Closes #254 

This PR adds two additional steps to the script that parses portal data:
- Create and export the expected metadata files into `s3_files/`
  - As such, I removed the original script to do this standalone  
- Copy files from S3 into `s3_files/reference_files/`

I made two additional changes uncovered during testing:
- I expanded the checks at the beginning to first check if an arg was provided before checking that it exists. I do think we want both checks since it seems helpful to have for external users.
- I updated how we deal with the overwrite argument; if at least one output directory exists and overwrite is true we need to remove directories before populating them to avoid `cp`/`mv` errors. So, I added code to remove directories with `rm -rf` in that section.

I also did a small README update, but we can expect most changes there will come in #211.


reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -183,10 +203,10 @@ input_zips <- list.files(
...
     \(x) {stringr::str_split_i(basename(x), "_", 1)}
   )
-stopifnot(
-  "The provided input directory does not contain all expected projects. Zip files for all projects listed in the `project-whitelist.txt` file should be present." = 
-    setequal(names(input_zips), expected_projects)
-)
+#stopifnot(
+#  "The provided input directory does not contain all expected projects. Zip files for all projects listed in the `project-whitelist.txt` file should be present." = 
+#    setequal(names(input_zips), expected_projects)
+#)
```
allyhawkins, line 209:
```suggestion
stopifnot(
  "The provided input directory does not contain all expected projects. Zip files for all projects listed in the `project-whitelist.txt` file should be present." = 
    setequal(names(input_zips), expected_projects)
)
```
I think we still want this here?

allyhawkins (approved):
LGTM

reproduce-figures/prepare-scpca-portal-data.R
sjspielman: yes! thank you!

Issue #245, opened 2025-04-01
Add reproducibility instructions to bulk analysis README


sjspielman:
> I think we could split up these instructions so instructions for analysis reproducibility are separate from this directory, and instead we can document that in the analysis README itself. This would be fine for reproducing figures since all the necessary files from the bulk analysis for this are either in the repo already or will be added to Zenodo (see https://github.com/AlexsLemonade/scpca-paper-figures/issues/244). What do you think of that organizational approach?

This sounds good to me.

_Originally posted by @allyhawkins in https://github.com/AlexsLemonade/scpca-paper-figures/pull/237#discussion_r2023308653_
            
---

As quoted above, we decided to split up instructions to reproduce figures vs. analysis. This issue tracks adding analysis instructions.

Edit: simultaneous with adding these instructions, we should upload to the Google Drive intended for Zenodo the intermediate data files generated in the analysis and saved but .gitignore'd in `data/`.

sjspielman:
Since consensus cell types are needed as input here, part of writing these instructions will depend on decisions following this comment: https://github.com/AlexsLemonade/scpca-paper-figures/issues/240#issuecomment-2775625842

sjspielman:
We can close this with #268.

PR #257, opened 2025-04-09
Sync the panglao references

sjspielman:
I started working on (edit, wrong link) #211 & #245, beginning with running the code all the way through. So, I found the first and hopefully last set of files we still need to grab - the panglao/cellassign marker gene references used in bulk analysis to make the gene sets. 

This PR adds that step to the data prep script. In addition,  I renamed the existing `reference_dir` variable so we can distinguish between s3_files vs bulk reference directories. I did make this change in VS Code, which ate up many spaces, so I recommend turning off whitespace!


allyhawkins (approved):
LGTM

PR #258, opened 2025-04-10
Account for SCEs without consensus cell types

sjspielman:
While testing code towards #211 & #245, I hit a bug in the consensus cell type functions at project SCPCP000020. This is a cell line project, so it didn't have the singler cell type column (which is temporarily coded up), but it certainly won't have a consensus annotation either! So, I updated this function to use `NA` values if the column isn't present. 

Additional small changes in the other script are present too:
- Fixed other small bug to change `scpca-data` to `scpca_data`
- I swapped an arg order to match the usage docs (makes it easier for me to work with code) and added a comment for future us.


allyhawkins (approved):
LGTM

PR #261, opened 2025-04-11
Update metadata parsing

### LINKED ISSUE #259: Update metadata parsing
While testing code towards #211 & #245, I encountered some issues/discrepancies with the metadata generation. In this issue, I'm laying out what I'm pretty sure is what we need to do to get the full data we need.

Currently, we rely on just the portal-wide metadata download, which isn't going to be comprehensive enough for a few reasons:
- It does not include an indicator for libraries with CITE-seq
- It does not include libraries that _only_ have bulk sequencing 

Therefore, we additionally need to obtain additional information from these files to complete the metadata:
- project-specific metadata files for CITE projects 
- bulk-specific metadata files

Moreover, there are four samples which are currently bulk-only that we need to include in the metadata files. Notably, for the sample metadata file that we need to recreate, we need information about `diagnosis`, `disease_timing`, and `is_cell_line`. However, none of this information appears to be obtainable from the portal:
- these samples are not in the portal-wide metadata download since they do not have single-cell libraries
- these samples are therefore not in the project-specific single-cell metadata either
- these samples _are_ in the bulk metadata tsvs, _but not with_ `diagnosis`, `disease_timing`, and `is_cell_line` columns to parse out.

Therefore, we may need to manually include these samples when creating the sample metadata tsv. 

(Note further, the BULK libraries from these samples are on the portal, but NOT any single-cell equivalent if it exists(?). So, it's not immediately clear to me if the fact that we have these annotated internally as `on_portal = TRUE` is correct?)

Bearing all this in mind, how should we update the parsing?

## Library metadata

- Combine information from bulk tsvs, citeseq project tsvs, and the portal-wide tsv
- Note that we cannot skip the portal-wide tsv here, since the project tsvs don't give all the information we need about about spatial libraries, but the portal-wide one does. 
  - For example, the sample `SCPCS000170` from `SCPCP000006` has _three_ rows in the portal-wide metadata, but only 1 row in the project-specific single cell representing the single-cell. So, the portal-wide seems necessary

## Sample data
- Use information from the portal-wide metadata, but _manually add in_ rows for the 4 samples which we can't get metadata for from the portal directly

sjspielman:
Closes #259 (if pursued!)

This PR updates our approach for preparing metadata files for reproducibility, according to the scheme I laid out in my opening comment of #259.

To get the full library information we need, I also save and parse bulk metadata TSVs and metadata TSVs for projects with CITE-seq, and handle cellhash samples a bit more carefully to ensure we get their technology information. To get the full sample information we need, I manually add in a table with the information we need.

The result seems to sufficiently match our metadata files! I used these new files, for example, to re-create Table S1 and Table S2. No diff for Table S2 🎉, but for Table S1 these versions of the metadata led to one diff from the table made with our internal metadata - the diagnosis for SCPCP000009 are in a different order, but are all the same. This is presumably just a row order thing and I'm not worried about it (unless we want to sort somehow in general) since the table has the same content.

At some point while I was here too, I made a few other changes (which can be cherry-picked out if we prefer):
 a bit of rearranging of directory definitions (eg the merged sce directory) to place them all higher in the script to create them all at once. I also fixed an exciting bug where I had called the library metadata file `"library-sample-metadata.tsv"`, which is...not correct.

### DIFF
--- a/reproduce-figures/prepare-scpca-portal-data.R
+++ b/reproduce-figures/prepare-scpca-portal-data.R
@@ -327,7 +327,7 @@ input_zips |>
 # Read input metadata file
 portal_metadata <- readr::read_tsv(opts$portal_metadata_path)
 sample_metadata_file <- file.path(opts$s3_files_dir, "scpca-sample-metadata.tsv")
-library_metadata_file <- file.path(opts$s3_files_dir, "library-sample-metadata.tsv")
+library_metadata_file <- file.path(opts$s3_files_dir, "scpca-library-metadata.tsv")
 
 # Create and export sample metadata table
  portal_metadata |>
--- a/reproduce-figures/prepare-scpca-portal-data.R
+++ b/reproduce-figures/prepare-scpca-portal-data.R
@@ -9,6 +9,8 @@
 #
 #####################################################################################
 ### CAUTION! YOU WILL NEED AT LEAST 170 GB OF AVAILABLE SPACE TO RUN THIS SCRIPT. ###
+###                                                                               ###
+### In addition, it will take roughly 90 minutes to run this script.              ###
 #####################################################################################
 #
 #
--- a/reproduce-figures/prepare-scpca-portal-data.R
+++ b/reproduce-figures/prepare-scpca-portal-data.R
@@ -353,11 +353,11 @@ sample_metadata_file <- file.path(opts$s3_files_dir, "scpca-sample-metadata.tsv"
 library_metadata_file <- file.path(opts$s3_files_dir, "scpca-library-metadata.tsv")
 
 # Create and export sample metadata table
-portal_metadata |>
-  dplyr::select(scpca_project_id, scpca_sample_id, diagnosis, disease_timing, is_cell_line) |>
-  # remove duplicate rows, which occur when there are multiple libraries per sample
-  dplyr::distinct() |>
-  readr::write_tsv(sample_metadata_file)
+prepare_sample_metadata(
+  portal_metadata,
+  sample_metadata_file
+)
+
 
 # Create and export library metadata table
 prepare_library_metadata(
--- a/reproduce-figures/prepare-scpca-portal-data.R
+++ b/reproduce-figures/prepare-scpca-portal-data.R
@@ -138,7 +138,14 @@ if ((dir.exists(opts$s3_files_dir) | dir.exists(opts$bulk_data_dir))) {
 consensus_files_dir   <- file.path(opts$s3_files_dir, "cell-type-consensus-results")
 celltype_results_dir  <- file.path(opts$s3_files_dir, "celltype_results")
 s3_files_reference_dir <- file.path(opts$s3_files_dir, "reference_files")
+merged_sce_dir <- file.path(opts$s3_files_dir, "SCPCP000003")
 bulk_reference_dir    <- file.path(opts$bulk_data_dir, "references")
+
+# these directories are for temporary file stored in scratch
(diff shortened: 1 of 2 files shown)


reproduce-figures/utils.R
```diff
@@ -75,3 +72,123 @@ prepare_gene_expression_tsv <- function(sce, marker_genes, output_tsv) {
...
+  library_metadata <- library_metadata |>
+    dplyr::bind_rows(cellhash_rows) |>
+    dplyr::select(-has_cellhash)
+  
+  # Define bulk and project metadata files
+  bulk_files <- list.files(
+    path = bulk_metadata_dir, 
+    pattern = "*_bulk_metadata\\.tsv", 
+    full.names = TRUE
+  )
```
allyhawkins, line 155: I think you should be able to search for this within the `project_scratch_dir` rather than copy files over to a bulk scratch directory first.

reproduce-figures/utils.R
```diff
@@ -75,3 +72,123 @@ prepare_gene_expression_tsv <- function(sce, marker_genes, output_tsv) {
...
+  citeseq_files <- list.files(
+    path = project_metadata_dir, 
+    pattern = "*_single_cell_metadata\\.tsv", 
+    full.names = TRUE
+  )
```
allyhawkins, line 160: I think same comment here. You can just provide one argument `project_scratch_dir` and grab the metadata files from there by recursively searching and then filtering based on the project ID being a CITE seq project.

reproduce-figures/prepare-scpca-portal-data.R
```diff
@@ -266,6 +277,20 @@ input_zips |>
...
+        fs::file_copy(project_bulk_tsv, bulk_metadata_scratch_dir, overwrite = TRUE)
+      }
+      
+      if (project_id %in% citeseq_projects) {
+        fs::file_copy(
+          file.path(project_scratch_dir, glue::glue("single_cell_metadata.tsv")), 
+          file.path(citeseq_metadata_scratch_dir, glue::glue("{project_id}_single_cell_metadata.tsv")), 
+          overwrite = TRUE
+        )
+      }
```
allyhawkins, line 293: Note some of my comments below, but I think you could remove this entirely and you don't need to copy metadata files from one scratch to the next scratch.

allyhawkins (commented):
This looks mostly good, I just have a comment about saving some copying of metadata files.

sjspielman:
So actually, the reason I am copying is to avoid hammering disk space. At the very end of the main purrr "loop", I have a line to remove the project scratch dir. This means that, after a given project is parsed, it's unzipped directory in scratch is removed:
https://github.com/AlexsLemonade/scpca-paper-figures/blob/a052a79f631865fe515fc87fcee2dc24ce6d316a/reproduce-figures/prepare-scpca-portal-data.R#L318-L320

If we don't so this, then the scratch directory will end up with all unzipped projects in it (which we'd remove in one line at the very end). This means users would need like 300-400 GB free space to run the code which is a heck of a lot and I wanted to avoid that burden. 

In any case, I'm not seeing a different strategy where we can avoid all the storage _and_ be able to not copy the files, so please cue me in if you see it?

allyhawkins:
> If we don't so this, then the scratch directory will end up with all unzipped projects in it (which we'd remove in one line at the very end). This means users would need like 300-400 GB free space to run the code which is a heck of a lot and I wanted to avoid that burden.

Ah that makes sense as to why you were copying then!

allyhawkins (approved):
LGTM

Issue #265, opened 2025-04-15
Test consensus cell type figures


sjspielman:
Following #253, we will need to test reproducibility of the consensus cell type figures when the input data was generated by `reproduce-figures/prepare-scpca-portal-data.R`.

sjspielman:
I have been able to successfully regenerate these figures using data downloaded from the portal. This can now be closed.

Issue #264, opened 2025-04-15
Regenerate FigS5A


sjspielman:
While testing reproducibility, I am getting the mirror of the UMAP for `figures/pngs/FigS5A_cellassign-umap.png`. I am getting this both with the processed RDS downloaded from the portal, and from the processed RDS in `s3://nextflow-ccdl-results/scpca-prod/results/`. Might be worth regenerating this and re-compiling S5?

Issue #263, opened 2025-04-15
Look into 3D script with merged UMAP


sjspielman:
While testing reproducibility updates, I ran into an issue where I'm not able to run [`scripts/Fig3D_merged-umaps.R`](https://github.com/AlexsLemonade/scpca-paper-figures/blob/540a3a22b6e6c02958e799340486234321cc5d91/scripts/Fig3D_merged-umaps.R) locally (M1 laptop, 32 GB RAM) due to memory. It seems the last time this figure panel was generated this past January in #113, and I don't think we've modified these objects on the portal side since. I specifically get the error, which I notably did not previously get!

```
Error: vector memory limit of 50.0 Gb reached, see mem.maxVSize()
Execution halted
```

One could bump this setting, but I'm a little wary of making changes like that that seem more like a user-specific need based on local architecture, and also because it's run for me before on this same computer.

So, I'm wondering if someone else could try running this script (from `main` is fine, doesn't need the new code) and tell me if you also get an error, and then maybe we compare notes... cc @allyhawkins!

sjspielman:
Want to note also, I ran this with `htop` and it slammed my swap memory (out of 7 gb) right away as it died, but RAM was fine overall.

allyhawkins:
It's because you are trying to use the full merged object rather than the smaller merged object that we have on S3. When I originally generated these figures I used the object in `scpca/processed` and not in prod. Perhaps we could try reading in that object and subsetting it? Although I'm pretty sure the problem is reading in that object in the first place.

allyhawkins:
Alternatively we can just not worry about it and note that you will need a lot of memory to create that figure from the merged object on the Portal.

sjspielman:
> Alternatively we can just not worry about it and note that you will need a lot of memory to create that figure from the merged object on the Portal.

Yeah, I mean I'm not sure what else we can do here since we can't re-create the UMAP scaling from just reading in those four libraries alone. 

But, it does seem like we need to update this compiled figure anyways since the UMAP I now get with the S3 object is a little different from what we had before, so I'll go ahead and do that at least. I will also add some comments about this script requiring more memory; I found a memory value where it works, so we can recommend that if their architecture allows.

sjspielman:
Hm, well, something is up actually with the UMAP layout. I think I know what it is but let's confirm so we head in the right direction! 

[...]

Issue #262, opened 2025-04-15
Update Figure 1 and manuscript numbers


sjspielman:
While testing reproducibility code, I was getting slightly different numbers in our `manuscript-numbers` tables and Figure 1 than we currently have - specifically, I'm getting 2 fewer samples:

- For the `manuscript-numbers/diagnosis-group-counts.tsv`, I am getting N=115 `Other Solid Tumors`, not N=117.
- For the `manuscript-numbers/disease-timing-counts.tsv`, I am getting:
  - N=12 `Progressive`, not N=13
  - N=520  `Initial diagnosis`, not N=521
- Looking closely at `Fig1_sample-summary.pdf`, I see two fewer neuroblastoma samples.

Happily, these are actually correct and this is not a new bug, but a bug we are catching! I confirmed these "new" numbers are the same numbers that the `main` branch here now returns. What seems to have happened was, _after_ we regenerated Figure 1 and number tsvs in #106, we made this update where 2 NB samples were incorrectly not marked cell line: https://github.com/AlexsLemonade/ScPCA-admin/commit/4f516bc94a4d3d17372a10248a42d6bdfa42dca6

Therefore, we will need to update some items here:
- [ ] `manuscript-numbers/diagnosis-group-counts.tsv`
- [ ] `manuscript-numbers/disease-timing-counts.tsv`
- [ ] `Fig1_sample-summary.pdf`
- [ ] Compiled figure 1

The first three items are taken care of by re-running `scripts/Fig1A_sample-disease-barchart.R`, but we of course need illustrator for the last one.

PR #266, opened 2025-04-15
Reproducibility testing miscellany

sjspielman:
Towards #211, #233, #262

This PR updates a couple items following testing of the `generate-figures-tables.sh` script from input data generated by `reproduce-figures/prepare-scpca-portal-data.R`. Note that other items not fixed here that still need to be addressed have all been opened as separate issues:
- #263
- #264
- #265

In addition, I have am now starting to fully test the bulk analysis updates by re-running that pipeline, so if I catch new errors there I will file a stacked PR to this one. 

This PR makes the following changes:

- I updated the `generate-figures-tables.sh` script to not delete and recreate `s3_files`. Instead, the script now expects that it exists and is populated with files. I therefore check if it exists, and if it does not I error out with a message to see the README for instructions on generating input data.
- I updated TSVs and the figure PDF to accomplish part of #262 (we'll still have to recompile figure 1 though)
- A few updates to `reproduce-figures/prepare-scpca-portal-data.R`:
	- Fiddling with paths lightly to get files copied into the right nested directories
	- The output directory for bulk references was incorrect; it's not inside `data/scpca_data/references`, but rather just `data/references`. I therefore updated how we do directories for bulk:
	  - The script now takes an argument for just the overall bulk data directory, and sets up to copy into the nested `scpca_data` and `references` directories from there.


allyhawkins (approved):
LGTM

Issue #267, opened 2025-04-15
Test pseudobulk analysis code


sjspielman:
The bulk analysis code relies on consensus cell types, which need to be added to the portal before we can fully test the code. at
I naturally remembered this halfway through testing today, so some testing has incidentally been done! I can confirm that, data set up via the `reproduce-figures` script leads to fully reproducible linear models ✅ 
But, we'll also want to confirm the ORA remains reproducible, which will need those consensus cell types.

Therefore, after #253, we'll need to test this pipeline again and address any bugs.

sjspielman:
Confirmed that results are fully reproducible with portal data and code runs as expected! This can now be closed.

PR #268, opened 2025-04-15
Analysis reproducibility fixes

### LINKED ISSUE #245: Add reproducibility instructions to bulk analysis README

sjspielman:
Stacked on #266 
Might close out #245

I began testing the bulk analysis code and identified some spots to fix which I've included in this PR. Note that more testing will be required as described in #267, so this is a step along the way!

This PR includes the following changes:

- The analysis script now expects that input data exists, and throws an error if input directories are not present
- The analysis script no longer runs the sync script, but I added instructions to README for what needs to be run to run this analysis. **Because of this**, I think we can maybe close #245? Imo, these README changes seems sufficient for _this_ README and the remaining instructions can be part of #211. Please let me know though if you think more docs would help here!
- Because the sync script produced the TSV file with the bulk sample/library ids which is used as input for other scripts, I ended up having to add that to the repo. It _only_ contains ids, but now it will be present when non-ccdl folks run the code.


allyhawkins (approved):
LGTM. I also think the readme is sufficient and you could close out #245

PR #273, opened 2025-04-16
Update panel 3D and code

### LINKED ISSUE #263: Look into 3D script with merged UMAP

sjspielman:
Closes #263 

This PR updates the panel 3D with merged UMAPs as discussed:

- We no longer sync the merged object, but the four relevant libraries
  - I updated both the internal and external data sync scripts to grab the processed rds files instead of the merged object
- The figure script has been updated to read in the four libraries, merge them (quick and dirty!), run multibatch PCA, and UMAP, and then plot. I took code straight from `scpca-nf` for this.  
  - Needed a couple more packages for this, so `renv.lock` was also updated


allyhawkins (approved):
LGTM

Issue #98, opened 2024-12-18
Identify potential datasets to compare bulk and single-cell


sjspielman:
This issue tracks identifying which samples we could pursue for a comparative cell type analysis between bulk & single-cell. We are seeking to analyze solid tumors with matched data: Which projects/samples are those?

For this, I'll comment on this issue a table (and probably tsv) of potential datasets and associated metadata of interest, e.g. number of cells.

PR #277, opened 2025-04-21
Fix sample numbers for modality bar chart Fig 1B

### LINKED ISSUE #276: Make sure numbers represented in modality bar chart and table are correct
I started reading through for my round robin edits and noticed that we have some discrepancies in sample numbers for the modality bar chart. It looks like we are over counting samples somewhere because we are showing 724 total samples in Fig 1B when we have 704 sample ids in the sample whitelist. I briefly started looking into it and noticed that we have an error in the script that creates Fig 1B with how we are counting the bulk and multiplexed samples, so that is probably part of it but I need to look into this further. 

This issue is to track figuring this out and updating the figure and supplemental table with the correct numbers.

allyhawkins:
Closes #276 

This PR makes an update to Figure 1B to ensure we are counting total samples and not total libraries. After going through the code, it looks like we were counting libraries for some modalities and samples for the other modalities and things just generally did not line up. I've cleaned up the code and ensured that we are counting samples for all modalities since that's what we say in the figure legend, results text, and the plot labels. Most of this involved adding in `unique` calls to make sure we didn't double count sample ids. 

An important note is that we have 704 samples included in our sample whitelist that should be accounted for here. The previous version had more than that in the single-cell and single-nuclei column, which was part of why I revisited this figure. These samples are broken down as follows: 

697 samples with single-cell/nuc data 
4 bulk only 
1 spatial only
2 bulk + spatial but no single-cell/nuc 

In this particular figure we show only samples that have single-cell/nuclei, so if samples were bulk or spatial only then they should not be included. 

### Single suspension 

For the "single suspension" column, the total should now be 699. There are two samples that have one single-cell library and one single-nuclei library, so those are now counted in both the single-cell count and the single-nuclei count adding 2 to the total 697 giving us 699. 

- I had to make a special adjustment to the code to account for these two samples that have one cell and one nuclei library since previously they were both getting counted twice in the single-cell group. 
- Side note that there are two samples that have two different libraries with different 10x kits so they were also being double counted. I made a fix to the code to remove that column before getting the totals and that solved that problem.

### Bulk 

We were missing a ton of bulk that were getting lost because we weren't properly dealing with the cell hashed samples. We should have 182 bulk samples, broken down as follows: 

142 solid tumor samples with bulk (as noted in https://github.com/AlexsLemonade/scpca-paper-figures/issues/98#issuecomment-2573164429) 
37 from SCPCP000003 
3 non-cancerous samples

The total in the figure should now be 176 since 6 of the bulk samples do not have associated single-cell/nuc data.

### CITE, Spatial and Cell hashing

These numbers should now reflect the total number of samples with this data and not the total number of libraries. 

### Other comments

- We now have less spatial than CITE-seq so the bars are not in order, but I think we want to keep bulk/spatial together and with cite/with cell hashing together. 
- I re-compiled Figure 1 after regenerating this figure. 
- I went through the code and the numbers for Table S1 and everything there looks correct, the only issue was with this panel. Adding the total number of samples in the samples column for S1 gives you 704 which is what we expect. The total bulk libraries is 182 and we don't have any scenarios where there is more than one bulk library per sample so this number should also reflect the total number of samples. 
- I'm going to file a PR to update the numbers in the text in `ScPCA-manuscript`. This will include figure legend updates to note the double-counted sample and that samples without associated single-cell or nuclei data are not shown.


jaclyn-taroni (approved):
👍🏻

PR #280, opened 2025-04-24
Update fig 2a alignment

### LINKED ISSUE #278: Update scpca-nf overview panel
While reading over the manuscript, I noticed that the text alignment in the `SCE Output Data` box in Fig 2A is a bit off: https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/figures/svgs/overview-scpca-fig2.svg

We should ensure this text is aligned and the update the figure 2 compilation.

sjspielman:
Closes #278 

This PR updates Figure 2, in particular by ensuring text and icons are properly aligned in Panel A. I updated the SVG panel separately first, and then did the compilation.


sjspielman:
ok it's _better_..!

sjspielman:
> Yea... I don't really see a difference here and when I open the new SVG in Illustrator the edges are cut off. I feel like this isn't worth trying to fix.

Yeah, we're shutting it down. It was fun while it lasted.

PR #282, opened 2025-04-25
Reproducibility updates for consensus cell types

### LINKED ISSUE #253: Update consensus cell type column in portal script
In #251, we're adding a script to parse portal data into files that can be used to reproduce figures/analysis. One aspect of this script is preparing consensus cell type TSVs by extracting this information from SCE libraries on the portal. Currently, however, consensus cell types are not on the portal. To get around this _temporarily_, the script currently grabs the `singler_celltype_annotation` column. 

We will need to make the following changes once consensus cell types are on the portal:
- Update the code to instead select grab the `consensus_celltype_annotation` column
- _But!_ to ensure the column is consistent with repo code, we should make sure it's named `consensus_annotation` in the output TSV

Once #251 is merged in, I will comment on this issue with the specific code line that needs to be updated.

Edit, here's the spot to update: https://github.com/AlexsLemonade/scpca-paper-figures/blob/a052a79f631865fe515fc87fcee2dc24ce6d316a/reproduce-figures/utils.R#L17-L19

sjspielman:
Closes #253 

This PR makes a couple changes to fully reproducibility of manuscript figures and tables:

- Per the original goal of #253, I updated the cell type column to parse consensus cell types and removed the singler placeholder
- Since the metadata file download is actually a zip file which contains a file `metadata.tsv`, I updated code and docs to account for this. Rather than reading a TSV, I check that it's a ZIP file, unzip it into a dedicated directory in `scratch`, and read in `metadata.tsv` from there.
- When running the dotplot code, I got a factor bug because the `unknown` level was repeated. I am pretty sure we could just safely remove the "unknown"` from this levels vector but I was nervous so instead just put it in a `unique()`! Either way the code is all running now :)

### DIFF
--- a/reproduce-figures/README.md
+++ b/reproduce-figures/README.md
@@ -28,17 +28,18 @@ Please refer to the [Introduction to `renv` vignette](https://rstudio.github.io/
 You will need to download the following files from the ScPCA Portal <https://scpca.alexslemonade.org/>:
 
 1. All ScPCA projects listed in the [project whitelist file](../sample-info/project-whitelist.txt)
-  * Use the project identifiers listed in the project whitelist file to navigate to each project page.
-For example, to navigate to `SCPCP000001`, use the URL: <https://scpca.alexslemonade.org/projects/SCPCP000001>.
+  * Use the project identifiers listed in the project whitelist file to navigate to each project page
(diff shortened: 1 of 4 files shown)


allyhawkins (approved):
LGTM

PR #287, opened 2025-12-03
Update consensus release date 

### LINKED ISSUE #228: Keep track of release for consensus cell types
I'm opening this issue just to note that we have used the `2024-11-25` OpenScPCA release for consensus cell types:

https://github.com/AlexsLemonade/scpca-paper-figures/blob/987c05bfb076620e84cb93adb45ca49225509028/scripts/figure_setup/sync-consensus-celltype-results.R#L30

These will not have changed in the more recent release (`2025-03-20`) but I think there is a non-zero potential for these to change following this investigation: https://github.com/AlexsLemonade/OpenScPCA-analysis/issues/1087

The purpose of this issue is to make sure we're using the correct results here if changes do happen.

sjspielman:
Closes #228 

Kicking off figure updates by updating where we get the figure data from! I added an option for the release date with default as current release to this setup script.

Edit - confirming I ran this locally, syntax is 👍

### DIFF
--- a/analysis/pseudobulk-bulk-prediction/README.md
+++ b/analysis/pseudobulk-bulk-prediction/README.md
@@ -25,7 +25,7 @@ To prepare and obtain all files needed for analysis input, please follow the ins
 ### Instructions for Data Lab members
 
 To prepare data for analysis, you will need to take the following steps.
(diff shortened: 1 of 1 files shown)


allyhawkins (approved):
LGTM

Issue #288, opened 2025-12-03
Determine if we need to remake pseudobulk panels


sjspielman:
The manuscript figures that display ORA results from the pseudobulk analysis may need to be updated given our refreshed consensus cell types.

This issue tracks running this analysis with current data release to determine if the figures need to be updated. If they do, we'll open separate issues to track making those changes.

PR #317, opened 2025-12-05
Refresh pseudobulk analysis files

sjspielman:
Related to https://github.com/AlexsLemonade/scpca-paper-figures/issues/288

This PR pushes the updated versions of result files from the pseudobulk analysis. I have re-run this using the newly released data, and we have tiny numerical changes based on small shifts in cell types, but _no_ changes to overall conclusions. So, the goal here is just to get the analysis files fully up to date with what will be going into the manuscript.


allyhawkins (approved):
LGTM

Issue #315, opened 2025-12-03
Keep track of script organization


sjspielman:
We're going to see a lot of script and file name changes since figure panels are moving around. I am opening this issue to refer to to keep us on track and less 😵‍💫 

Here is a table of scripts whose names/contents we anticipate needing to change:

| Script                                    | Revision panel destination                                                             | Tracked? |
| ----------------------------------------- | -------------------------------------------------------------------------------------- | -------- |
| `Fig3D_merged-umaps.R`                    | S3D                                                                                   | \#310    |
| `Fig4B_singler-cellassign-heatmap.R`      | 4C                                                                                     | -        |
| `Fig5A-S6_consensus-cell-type-dotplots.R` | 4A and S5 (all panels)                                                                              | -        |
| `Fig5B-C_brain-barplots.R`                | 4B-C                                                                                     | -        |
| `Fig6-FigS8_bulk-analysis.R`              | 6 (same) and **S7** (all panels)                                                                    | \#314    |
| `FigS4_celldex-ref-comparison.R`          | None - panel being removed                                                             | -        |
| `FigS5A_cellassign-justification.R`       | None - panel being removed                                                             | -        |
| `FigS5B_submitter-celltypes-heatmap.R`    | TBD; panel is either being removed or heavily updated for use in an updated **Fig S4** | -        |
| `FigS7_consensus-bar-plots.R`             | S6  (all panels)                                                                                   | -        |

We do not anticipate that these panels will change:

- `Fig1A_sample-disease-barchart.R`
- `Fig1B_modality-barchart.R`
- `Fig2B-F_qc-plots.R`
- `FigS1A_memory-time-comparison.R`
- `FigS1B-D_method-metrics-comparison.R`
- `FigS2B-D_adt-plots.R`

For more context, see these slab note and the linked google slides from discussion: https://data-lab-knowledge.slab.com/posts/2025-12-03-sc-pca-manuscript-r-1-figure-revisions-discussion-4v73s0p0

sjspielman:
We're going to do this as a real stand-alone issue to file a PR for, as discussed: https://github.com/AlexsLemonade/scpca-paper-figures/pull/325#issuecomment-3634268825

PR #325, opened 2025-12-08
Add script for NB UMAP panels

### LINKED ISSUE #318: Create NB UMAP panels
This issue tracks adding a script to create NB panels with UMAPs, specifically panels A & B as sketched out here: https://docs.google.com/presentation/d/1P7Y32XXWFuEcw2IFCDnDPr9Cs3Zs_QJpS-Ry9wYGKoE/edit?slide=id.g3ad10d43437_0_45#slide=id.g3ad10d43437_0_45
These are currently slotted to be the new figure 5. 

In #289, we assessed the feasibility of making these figures and concluded:

- For panel A, we want to use cell types collapsed into categories malignant, immune, endo, fibro, schwann, and unknown (unknown + openscpca-excluded) 
- For panel B, we can just color by total cnv fine. The non-colored blobs with `NA` represent samples (n=16) which weren't processed through inferCNV due to insufficient reference cells

sjspielman:
Closes #318 

This PR adds a script and two new pdfs for the 5A and 5B UMAPs.

- For the celltype panel, I used lowercase cell type names and labels to match how we represent this information in other figures. I'd appreciate some feedback on the legend title though, which is currently `OpenScPCA broad cell type annotation` - how does this feel to you? Any suggested tweaks? 
Also, let me know if we prefer that I make a separate TSV file for the palette; since this is the only time that palette will be used, I left it as a vector in the script but we can definitely pull it out!
- I don't have much to say about the inferCNV umap, it's pretty straightforward I think?
- I also added the new script to `generate-figures-tables.sh`, and left various comment organizational breadcrumbs for future-us when we rearranging things for #315 

Requesting @jaclyn-taroni for any more figure-related feedback, and @allyhawkins for figure feedback+code review


scripts/Fig5A-B_nb-umap-panels.R
```diff
@@ -0,0 +1,120 @@
...
+      "openscpca_celltype_annotation"
+    ),
+    use.dimred = "UMAP"
+  ) |>
+  # remove rownames
+  tibble::as_tibble()
+
+# define output files
+output_panel_a_file <- here::here("figures", "pdfs", "Fig5A-umap-celltypes.pdf")
+output_panel_b_file <- here::here("figures", "pdfs", "Fig5B-umap-infercnv.pdf")
```
allyhawkins:
```suggestion
output_panel_a_file <- here::here("figures", "pdfs", "Fig5A_umap-celltypes.pdf")
output_panel_b_file <- here::here("figures", "pdfs", "Fig5B_umap-infercnv.pdf")
```
Keeping with the previous naming format, we use `_` between the figure panel and description.

scripts/Fig5A-B_nb-umap-panels.R
```diff
@@ -0,0 +1,120 @@
...
+
+# define nbatlas condensed palette
+cell_palette <- c(
+  "malignant cell" = "rosybrown4",
+  "immune cell" = "darkgreen",
+  "endothelial cell" = "#FC1CBF",
+  "fibroblast" = "#B10DA1",
+  "schwann cell" = "chocolate1",
+  "unknown" = "gray70" # darker than validation palette since we have some standalone unknown clumps
+)
```
allyhawkins: I know it's only being used here, but I find it helpful to have the palettes as tsvs in the same folder so that if we need to add additional figures and re-use or modify this palette it's easy to find and use.

allyhawkins:
> > Ok, I will just tackle #315 as a "standalone" issue and rename things first, and then circle back here. Something I haven't been sure about is what to do about the PDFs though... for the ones that we don't expect to change at all, I will rename them, but for ones that may/will change (e.g. due to consensus updates), maybe for now I remove them? @allyhawkins any preference as my presumed reviewer?
> 
> Why don't we move any outdated pdfs to an old folder that we can remove later if we want? It might be helpful to see how things are changing between versions with consensus for responding to reviewers.

Also, I don't think this needs to be done before this PR goes in necessarily, I just think it would be helpful to do before we start making a ton of changes so before the next figure PR is good!

scripts/Fig5A-B_nb-umap-panels.R
```diff
@@ -0,0 +1,117 @@
...
+    cell_category = stringr::str_to_lower(cell_category),
+    # determine factor order based on frequency, with Unknown last
+    cell_category = forcats::fct_infreq(cell_category),
+    cell_category = forcats::fct_relevel(cell_category, "unknown", after = Inf)
+  )
+
+# prepare palette
+nb_palette <- readr::read_tsv(palette_file)
+cell_palette <- nb_palette$color
+names(cell_palette) <- nb_palette$celltype
```
allyhawkins:
```suggestion
cell_palette <- readr::read_tsv(palette_file) |> 
  tibble::deframe()
```
Simplifying this!

Issue #291, opened 2025-12-03
Update color palette for finer immune expression


sjspielman:
This issue generally tracks updating our consensus cell type color palette to include finer-grained T cells specifically, which are generally "blue" colors. 

Noting I'm opening this as a separate issue for overall coordination since this blocks a couple figures down the line.

sjspielman:
This has been completed as part of #329

PR #329, opened 2025-12-10
4D immune cell dotplot

### LINKED ISSUE #293: Investigate making a dot plot with finer grained immune cell types for Fig 4D
We would like to create a new dotplot that focuses in on marker genes for finer grained immune cell types for the samples in Fig 4. 

We need to first make a list of all the immune consensus cell types that are found in the samples used in Fig 4 and see if we have marker genes for all those cell types. We have compiled a list of marker genes in [`OpenScPCA-analysis` for any cell types that were available in CellMarker2.0](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/validation-markers.tsv). 

If there are cell types that are missing, we should look at [cellxgene census](https://cellxgene.cziscience.com/cellguide) or using [Elicit](https://elicit.com/) to curate a marker gene list. 

Once we know how many cell types we have and identify marker genes then we can make the dotplot and figure out colors 😬

allyhawkins:
Closes #293 

This PR adds a dot plot with all immune consensus cell types from Brain and CNS samples. This includes the script to generate the pdf, the actual pdf, and an updated immune cell type palette to include new cell types that were not previously accounted for. 

* One key thing to note here is that I had a slight error when getting the marker genes to include in my mock up that I made showing the lineage specific marker genes (see https://github.com/AlexsLemonade/scpca-paper-figures/issues/293#issuecomment-3634484704). Instead of getting genes that were specific to the lineage, I was getting genes that were specific to only the cell types included in the plot (all immune cells), which I think is also a reasonable approach.

I think we should keep this because if I update it to keep all lineage specific genes the plot still looks bonkers to me: 
Screenshot 2025-12-10 at 3 02 40 PM

* Within the script itself, I ended up copying the code to make the marker gene dot plot and making modifications where necessary. There was a fair amount of nuanced items, like ordering and grouping in the main function we have that didn't quite fit with this implementation. Mainly, we infer orders in that function using specific columns in a way that doesn't apply here and after spending some time trying to tease out handling the factors for different implementations differently, I felt it wasn't worth the effort. 

* I pulled out the actual plotting part of the function into a helper function that I can use here, which did add some simplification. 

* I modified the existing immune cell palette to include some new shades for the immune cells that are now present here. This should mean we can remake the bar plots for 4C for the brain immune easily.

### DIFF
--- a/generate-figures-tables.sh
+++ b/generate-figures-tables.sh
@@ -61,6 +61,9 @@ Rscript ${script_dir}/Fig4A-S5_consensus-cell-type-dotplots.R
 # https://github.com/AlexsLemonade/scpca-paper-figures/issues/297
 Rscript ${script_dir}/Fig4B-C_brain-barplots.R
 
+# Figure 4D 
+Rscript ${script_dir}/Fig4D_immune-only-dotplot.R
+
 # Figure 5A-B
 Rscript ${script_dir}/Fig5A-B_nb-umap-panels.R.R
 


sjspielman:
I had a quick look but will do a full/actual review tomorrow.. for now just 1 FYI - looks like you also get to close #291 🎉

scripts/Fig4D_immune-only-dotplot.R
```diff
@@ -0,0 +1,205 @@
...
+  dplyr::summarize(
+    .by = c("consensus_annotation", "ensembl_gene_id", "validation_group_annotation"),
+    detected_count = sum(detected),
+    mean_exp = mean(logcounts)
+  ) |>
+  # add in validation group for marker genes
+  # this includes all possible marker genes and all possible validation group assignments
+  dplyr::left_join(markers_df, by = c("ensembl_gene_id", "validation_group_annotation"), relationship = "many-to-many") |> # add total cells
+  dplyr::left_join(total_cells_df, by = c("consensus_annotation")) |>
+  # for plotting we're only going to look at any cell types with > 50 cells otherwise these plots can get wild
```
sjspielman: i approve leaving this comment 😄

sjspielman (commented):
LGTM! I ran the script locally and the plot looks good (good luck to future us compiling though, can't be helped with these 😬).

I left a couple small spacing comments and a bigger comment about loading renv, but we should get this into `generate-figures-tables.sh`, too. Might want to wait til #330 goes in to reduce conflicts? That said, conflicts shouldn't be _too_ bed.

allyhawkins:
@sjspielman I updated the spacing you found, added renv, and added it to the main figure generation script. This should be ready for another look.

PR #333, opened 2025-12-11
Add new fig 1C with Cell Browser

### LINKED ISSUE #298: Update Figure 1C to include cell browser
This issue tracks incorporating an "inset" of a cellbrowser view into Figure 1C. The project card aspect itself may need to also be resized to accommodate all this, which will involve some trial and error.

Google slides context: https://docs.google.com/presentation/d/1P7Y32XXWFuEcw2IFCDnDPr9Cs3Zs_QJpS-Ry9wYGKoE/edit?slide=id.p#slide=id.p

jashapiro:
closes #298
I changed the project card to SCPCP00004, and included a sample with a nice diversity of cell types for the new Cell Browser screenshot.

The updates here will also require changes in the figure legend, which are tracked in https://github.com/AlexsLemonade/ScPCA-manuscript/issues/204

I moved the old version to v1_figures directory.


jashapiro:
> Oh dear...fig looks fine, but I cannot approve this since it looks like there's a bug on the portal side. There are _definitely_ 42 samples in this project, not 40. We indeed see this reflected on the project page sample summary table https://scpca.alexslemonade.org/projects/SCPCP000004, but the top of the project card incorrectly says 40.
> 
> I'll file an issue in `scpca-portal` and note it in slack, too.

There are 40 samples in the sample list, and I see 40 on the page at https://scpca.alexslemonade.org/projects/SCPCP000004. There are however, 42 _libraries_ as there are two samples with two libraries.

Issue #334, opened 2025-12-12
UMAP of consensus cell types with total inferCNV for S4


allyhawkins:
As panel B in S4, we want to include side by side UMAPs of consensus cell types and total inferCNV for SCPCS000049/SCPCL000049. This issue tracks creating that panel. I think we might be able to use the same script we use for #299 and then use patchwork or cowplot to stack the faceted UMAP in panel A on top of this panel, but depends on how that looks. 

See https://github.com/AlexsLemonade/scpca-paper-figures/issues/299#issuecomment-3647235991 for the original context.

PR #338, opened 2025-12-16
Supplemental figure showing T cell types across automated methods

### LINKED ISSUE #299: Make a UMAP of T cell types annotated with all four automated labels
We want to make a UMAP for each of the four automated labels - SingleR, CellAssign, SCimilarity, and consensus. If we showed all cell types this could get messy quickly because of the different cell type labels used by each method. This issue tracks trying to make the UMAP with just T cell types and all other cells greyed out. We need to see if any samples that would work for this (maybe AML or a solid tumor?). 

Note that the figure ideally includes an inferCNV UMAP too so in a perfect world we have a sample with T cell types and inferCNV results. 

Google slide for context: https://docs.google.com/presentation/d/1P7Y32XXWFuEcw2IFCDnDPr9Cs3Zs_QJpS-Ry9wYGKoE/edit?slide=id.g3ad10d43437_0_100#slide=id.g3ad10d43437_0_100

allyhawkins:
Closes #299 

Here I'm adding the faceted UMAP looking at the top T cell types for each automated method in a single sample. The sample being shown is SCPCL000049/SCPCL000049. This has been a little bit of a journey because we want to get the top cell types for each method individually and then only show T cell types in the facets but all cells in the actual UMAPs themselves. Also each method has their own top T cell types to show.  

Ultimately, it turned out to be easier than I had originally made it, but the remaining difficulty is getting the x and y axis labels to cooperate. I think it's good enough and then alignment can be further adjusted in illustrator. I can only fight with patchwork and ggplot so much.

I also had to make sure I was capturing all T cell types, so I added a list of all possible T cell types across all methods in this sample to our `sample-info` folder. 

As part of this PR, I also added the code to sync the data for this sample to our `sync-data.R` script. 

I plan to tackle #334 next and have some code that I used for testing that out. I left that in right now but just commented it out. I think it probably makes sense to have all of that code in this script since those two plots are much more straight forward and come from the same sample and setup code. If you agree, I'll also update the name of the script here to reflect that. 

Requesting @jaclyn-taroni for figure review and @sjspielman for code.


scripts/FigS4_t-cell-umap.R
```diff
@@ -0,0 +1,172 @@
...
+  tibble::as_tibble()
+
+# Faceted UMAP - panel A -------------------------------------------------------
+
+# list of cell type columns to lump and plot
+celltype_columns <- c(
+  "Consensus" = "consensus_celltype_annotation",
+  "SingleR" = "singler_celltype_annotation",
+  "CellAssign" = "cellassign_celltype_annotation",
+  "SCimilarity"= "scimilarity_celltype_annotation"
```
sjspielman: Let's sync up here: for my pending heatmap, I was ordering these in the plot by alphabetical order (cellassign, scimilarity, singler). I think that makes more sense for the paper? This order is more how the sausage was made.

sample-info/SCPCS000049_T-cell-types.txt
sjspielman: I do think we want a script in `scripts/figure_setup` to generate this file, for transparency. I expect that script to be a whopping like 3-4 lines.

allyhawkins: So I made this manually be exporting all cell types and then removing any non T cell types. Alternatively we could just grab all T cell types in the code with `stringr::str_detect("T ")`. That just made me nervous that we would accidentally get something that isn't a T cell.

sjspielman: >  Alternatively we could just grab all T cell types in the code with stringr::str_detect("T "). That just made me nervous that we would accidentally get something that isn't a T cell.

This seems like it's fine to me? I get the same list you get with this code where I'm checking for both "T " and "T-".

```
library(SingleCellExperiment)

sce <- readRDS(
  here::here(
    "s3_files", 
    "SCPCS000049", 
    "SCPCL000049_processed.rds"
  )
)

colnames <- c(
  "singler_celltype_annotation",
  "cellassign_celltype_annotation", 
  "scimilarity_celltype_annotation", 
  "consensus_celltype_annotation"
)
all_celltypes <- colnames |>
  purrr::map(
    \(col) unique(sce[[col]])
  ) |>
  purrr::reduce(union)

all_celltypes[stringr::str_detect(all_celltypes,"T[- ]")] |> sort() 
```

jaclyn-taroni (approved):
👍🏻  I will warn you that there is some chance I change my mind about order when we start writing things, but hopefully not!

scripts/FigS4_celltype-method-umaps.R
```diff
@@ -71,7 +67,7 @@ all_plot_list <- celltype_columns |>
       dplyr::mutate(
         t_cell_group = dplyr::if_else(
           # keep T cells with the original annotation and leave all others as NA
-          !!sym(celltype_column) %in% t_cell_types,
+          stringr::str_detect(!!sym(celltype_column), "T[- ]"),
```
sjspielman: 👍

PR #340, opened 2025-12-17
Fig S4B UMAP with consensus cell types and inferCNV

### LINKED ISSUE #334: UMAP of consensus cell types with total inferCNV for S4

allyhawkins:
Closes #334 

Here I'm adding panel B for S4 which will show the top consensus cell types and the total CNV for the same sample where we show the T cell automated cell type annotations added in #338. 

I chose to show the top 7 cell types and then group all others into "All remaining cell types" like we do in the reports. I added the total number of cells to the labels so you can see that also. 

I think the biggest question here is around the choice of colors I use for the consensus cell types. I chose to just use the Dark2 palette since that's what we use in other places where we show consensus cell types. Are we okay with this? We do have some colors in our validation and immune palettes that we could use for everything but the OPC and neuron associated cell, but they are all shades of green/yellow and were kind of hard to distinguish in a single UMAP so I went with Dark2. 

Requesting @jaclyn-taroni for figure review and @sjspielman for code.

### DIFF
--- a/scripts/FigS4_celltype-method-umaps.R
+++ b/scripts/FigS4_celltype-method-umaps.R
@@ -152,7 +152,6 @@ ggsave(tcell_umap_file, combined_plot, width = 10, height = 10)
 umap_df <- umap_df |> 
   dplyr::mutate(
     consensus_lumped = forcats::fct_lump_n(consensus_celltype_annotation, n = 7, other_level = "All remaining cell types") |> 
-      #stringr::str_wrap(25) |> # wrap the long names
       forcats::fct_infreq() |> 
       forcats::fct_relevel("Unknown", "All remaining cell types", after = Inf)
   )


scripts/FigS4_celltype-method-umaps.R
```diff
@@ -146,35 +148,57 @@ ggsave(tcell_umap_file, combined_plot, width = 10, height = 10)
...
-#     y = "UMAP2",
-#     color = "Total CNV"
-#   ) +
-#   theme(legend.position = "bottom")
-
+# lump consensus cell types 
+umap_df <- umap_df |> 
+  dplyr::mutate(
+    consensus_lumped = forcats::fct_lump_n(consensus_celltype_annotation, n = 7, other_level = "All remaining cell types") |> 
+      #stringr::str_wrap(25) |> # wrap the long names
```
sjspielman: I gather this turned out not to be needed

```suggestion
```

I don't think labels are too long with the `(n = {n})` added in either, so this all seems fine.

PR #356, opened 2026-01-06
Compiled Fig4

### LINKED ISSUE #307: Recompile figure 4
This issue tracks updating the compiled figure 4, in which we currently are planning to include a new panel D which will be a dotplot of finer-grained immune cell types.

Google slides context: https://docs.google.com/presentation/d/1P7Y32XXWFuEcw2IFCDnDPr9Cs3Zs_QJpS-Ry9wYGKoE/edit?slide=id.g3ad10d43437_0_34#slide=id.g3ad10d43437_0_34

sjspielman:
Closes #307 

This PR updates the fig4 compilation:
A: brain/cns dotplot
B: brain/cns cell types barplot
C: brain/cns immune cell types barplot
D: brain/cns immune dotplot

I bumped the text size as much as I could, but inevitably a lot remains small because of the dotplots. But, now that we're capped at 10 marker genes, there's less text than there could be!


sjspielman:
> This looks good, but before I approve, should we also cap the number of marker genes for the immune cell plot in D? That would just affect CD8-positive, a/b memory T cell which has 14 genes. I feel like we should to be consistent across all the plots?

Thanks for catching this - something happened since this was meant to be fixed in https://github.com/AlexsLemonade/scpca-paper-figures/pull/354. Please hold!

sjspielman:
@allyhawkins the right panel version is there now! No code changes were needed (except a little resize) since those happened in #354 but apparently I didn't get the updated panel in there properly!

PR #359, opened 2026-01-06
Diagram for consensus cell type annotation

### LINKED ISSUE #301: Make a diagram explaining how consensus cell type assignment works
This issue tracks making a diagram for Figure 3B outlining the approach for assigning consensus cell types. One idea is to use T cell types as an example to show how LCAs are used to assign the final consensus cell type label. 

We should start by drawing some options and then we can go from there.

allyhawkins:
Closes #301 

Here I'm adding the diagram to use for 3B to explain consensus cell types following the discussion in #301. Are we okay with these colors?


sjspielman (approved):
👍

PR #363, opened 2026-01-07
Recover Figure 2

allyhawkins:
Not sure what happened, but this fixes the issues with figure 2 noted in https://github.com/AlexsLemonade/scpca-paper-figures/pull/355#issuecomment-3720328167


jaclyn-taroni (approved):
🚀

PR #366, opened 2026-01-09
Update reproducibility code and docs for portal-wide downloads

### LINKED ISSUE #286: Update reproducibility docs to include info about ScPCAr
In #252, we describe how we might benefit from a script that supports downloading projects. However, at this point, we now have a working implementation of a package wrapping the API: https://github.com/AlexsLemonade/ScPCAr/

Rather than providing a script, we should update reproducibility docs to point users to the API and describe how they could use it to set up files locally to run the code.

### LINKED ISSUE #323: Add NB datasets to reproduce-figures script
Related to #322

We'll need to get the NB library & merged object listed in #322 into the `reproduce-figures/prepare-scpca-data-portal.R` script as well.

sjspielman:
Closes #286  (except I didn't actually consider `ScPCAr` in the end - I went with portal-wide downloads instead, much easier)
Closes #323 

This PR updates and simplifies reproducibility code/docs now that we have an option for the portal-wide download! 
I have tested these changes, and I was able to run the updated script included here and then `generate-figures-tables.sh` to successfully regenerate figures and tables ✅ 

The script changes:
- Previously, users would have had to download each project's ZIP independently. We then `purrr::walk`ed over each project and unzipped it into a scratch directory.
- With the portal-wide download, we no longer need a scratch directory because everything in the download is uncompressed. This simplifies processing a lot! Also because of the download structure, the checks at the beginning of the script got updated a bit too
- As such, users now need to download 3 items, each of which they should uncompress and input the local directory for to the script
  - portal-wide sce download
  - portal-wide metadata download
  - project SCPCP000004 as merged (for NB figures; #323. the script now includes additional code for grabbing the NB files too)
- Note that we do still need the portal-wide metadata download, since we have those 4 samples which are bulk only and this is the only TSV which has the full metadata for those samples; it's not included in the relevant bulk metadata tsv's.
- I removed the code to sync the SingleR model files and the t2g file since we no longer use those for a figure. I went ahead and ripped this out of our own reference syncing script too, but can restore if you think that's important for posterity/old figures.

### DIFF
--- a/reproduce-figures/README.md
+++ b/reproduce-figures/README.md
@@ -75,9 +75,9 @@ This script will run for roughly 75 minutes, and it will create three directorie
 Note that this script will not overwrite existing output directories without the `--overwrite` flag.
 For more options, use the `--help` flag.
 
-After running the script, you can remove all downloaded portal ZIP files and directories as desired.
+After running the script, you can remove all downloaded portal ZIP files and directories.
 
-### Step 5: Run repository code
+### Step 4: Run repository code
 
 Once `prepare-scpca-portal-data.R` script has successfully completed running, you will be able to run the following repository scripts to reproduce figures and results:
 


reproduce-figures/README.md
```diff
@@ -3,17 +3,16 @@ It contains the following contents:
...
-  * **CAUTION: Running this script requires at least 170 GB of free storage space**
-  * This script uses functions provided in `utils.R`
-  * This script also uses the `./scratch` directory during file processing
+  * This script uses functions provided in `utils.R`.
 * TSV files, `FigS1B-gene-expression-data.tsv` and `FigS1C_FigS1D-cell-metrics-data.tsv`, which are used as input to the script `../scripts/FigS1B-D_method-metrics-comparison.R` to create Figures S1-D.
-  * Note that this script expects these TSV files to be present in this directory; please do not move them
+  * Note that this script expects these TSV files to be present in this directory; please do not move them.
 
 
+**CAUTION: You will need at least 300 GB of free storage space for downloaded ScCPA Portal data and associated data preparation.**
```
allyhawkins: Wow

reproduce-figures/README.md
```diff
@@ -25,41 +24,44 @@ Please refer to the [Introduction to `renv` vignette](https://rstudio.github.io/
...
+    You should unzip this file, which will create a directory `portal-wide_single-cell-experiment_<accessed-date>`.
+    Note the location of this directory for script input.
+    You can delete the ZIP file at this point if you prefer.
+4. Next, navigate to the Portal project page for `SCPCP000004`: <https://scpca.alexslemonade.org/projects/SCPCP000001>
+  * Click `Download Now` to download this project as a single merged object, selecting the following settings:
+    * Modality: `Single-cell`
+    * Data Format: `SingleCellExperiment (R)`
+    * **Check the box to** `Merge samples into 1 object`
+  * This will download a ZIP file called `SCPCP000004_single-cell-experiment_<accessed-date>.zip`.
+    You should unzip this file, which will create a directory `portal-wide_single-cell-experiment_<accessed-date>`.
```
allyhawkins:
```suggestion
    You should unzip this file, which will create a directory `SCPCP000004_single-cell-experiment_<accessed-date>`.
```

reproduce-figures/README.md
sjspielman: Worth noting that this is really a worst case scenario if they have everything downloaded zipped AND unzipped. If they delete the zip files after compressing, they will only need half of this!
Do you think I should expand on that?

Issue #384, opened 2026-01-30
Regenerate stacked bar plots


allyhawkins:
We need to regenerate the bar plots showing consensus cell types for Fig 4B, 4C and all of S6.

PR #397, opened 2026-02-05
Regenerate 4B and 4C - stacked barplots for brain samples

allyhawkins:
Towards #384 

This updates the stacked bar plots with brain and CNS samples that are part of figure 4. There were a few more changes here than I originally thought so I'm starting with these two and then I'll regenerate the rest in a separate PR. 

I updated the first bar plot that shows the broad cell type groups and double-checked that we don't have any additional validation colors that we need to add. That one just meant re-running the script. 

For 4C, it looks like we never updated the list of T and myeloid cell types to include in the plot when we updated consensus to incorporate `SCimilarity`. We previously had hard-coded a list of cell types to include in each group, which isn't sustainable now that we have a lot more possible immune cell types. Also, this [file that we were using to pull in immune cell types has not been updated](https://github.com/AlexsLemonade/OpenScPCA-analysis/blob/main/analyses/cell-type-consensus/references/consensus-immune-cell-types.tsv), so I'll file an issue in `OpenScPCA-analysis` to update it. I'm fairly certain this is the only place we used it though and with the changes here we don't need it anymore. 

Instead of pulling the list of immune cell types and listing out all the T/myeloid cell types, I added a list of validation groups that contain immune cell types. I then pulled out the T cell and myeloid cell groups and used that to grab the individual consensus cell type annotations. I think this will be easier to update if we ever have more immune cell groups than having to list all consensus cell types. 

Now that we have a lot more possible T and myeloid cell types, I implemented a threshold so that we don't show any libraries with < 1% immune cells and any types with < 1000 total cells across all libraries are grouped with "other". We will need to update the legend to reflect this since previously we only mentioned removing `granulocytes`. In reality, the cell types that get removed all have less than 200 total cells, but 1000 seemed like a good number in case we add more samples and have to regenerate this. I would like to avoid having to show 20+ cell types. 

I added a little message to the code to tell us what cell types get removed, but I think saying we remove cell types with < 1000 total cells in the figure legend is reasonable enough.

Finally, we have new cell types in the figure and one that we didn't have a color for before, `tissue-resident macrophage`. I added a color for that and also cleaned up some of the other colors that do get shown together to make them slightly more distinct when viewed together on a barplot. Because one of the colors I changed is also in the color bar for the dot plot in 4D, I re-ran that script to regenerate the plot. 

Here's the updated 4C plot: 
[Fig4C_brain-barchart-immune.pdf](https://github.com/user-attachments/files/25110738/Fig4C_brain-barchart-immune.pdf)

### DIFF
--- a/scripts/Fig4B-C_brain-barplots.R
+++ b/scripts/Fig4B-C_brain-barplots.R
@@ -106,6 +106,8 @@ ggsave(output_4b_pdf_file, plot = hgg_lgg_barplot, width = 8, height = 8)
  # Prep and plot 4C -----------------------------------------------------
 
 # get all possible immune cell types
+# use the validation groups from consensus-validation-groups.tsv that have immune cells
+# don't include HSPCs since they are borderline "immune"
 all_immune_groups <- c(
   "B cell", 
   "T cell", 


scripts/utils/consensus-celltype-plotting-functions.R
```diff
@@ -378,6 +380,31 @@ create_immune_celltype_summary <- function(
...
+    dplyr::group_by(immune_celltype_group) |> 
+    dplyr::summarise(total_cells = sum(total_cells_per_annotation)) |> 
+    dplyr::arrange(desc(total_cells))
+  
+  # filter 
+  low_count_celltypes <- all_celltype_totals |> 
+    dplyr::filter(total_cells < minimum_cell_count) |> 
+    dplyr::pull(immune_celltype_group)
+  
+  if(length(low_count_celltypes) > 0){
```
sjspielman:
```suggestion
  if (length(low_count_celltypes) > 0) {
```

PR #398, opened 2026-02-05
Update all consensus cell type dot plots

### LINKED ISSUE #383: Regenerate consensus cell type dot plots
We need to regenerate the dot plots for 4A and S5 after updating the consensus cell types.

allyhawkins:
Closes #383 

Here I'm updating all the dot plots with the new consensus cell types. When doing this, I wanted to have an easy way to identify cell types that were not being shown due to lack of marker genes and tell us if we were missing any colors for the annotation bar at the top. I added a few checks with messages to print that information out for future us. 

We don't have any validation groups that are missing colors in these plots and the only missing groups without marker genes are `pigment cell` in Sarcoma and `pigment cell` and `kidney cell` in other tumors. We already report these as missing in the legend so nothing to update there.


sjspielman (commented):
Figures themselves look good to me, but left a comment about rearranging code just a tad

allyhawkins:
I updated the code to pull it out into a separate function. This should be ready for another look.

sjspielman (approved):
thanks, lgtm!

PR #400, opened 2026-02-06
Regenerate stacked bar plots for remaining groups

### LINKED ISSUE #380: Update validation palette to include the remaining 4 colors
As noted in https://github.com/AlexsLemonade/scpca-nf/issues/1159, we are missing 4 groups from our validation palette that could theoretically appear in our dataset. It's possible that we don't need colors for all the groups, but I think it's reasonable to just add 4 colors to the palette in case they do show up so that the palette includes the full set of validation groups possible.

### LINKED ISSUE #384: Regenerate stacked bar plots

allyhawkins:
Closes #384
Closes #380 

I re-ran the script to make the new stacked bar plots for S6 and am adding the new figures here. We have a nice little check in that function already that reports if there are missing colors and everything was able to generate without an error so we don't need to update any palettes here. We definitely have more annotated cell types if you compare to previous versions so that's fun! 

New files:
[FigS6A_leukemia-barplot.pdf](https://github.com/user-attachments/files/25134285/FigS6A_leukemia-barplot.pdf)
[FigS6B_sarcoma-barplot.pdf](https://github.com/user-attachments/files/25134286/FigS6B_sarcoma-barplot.pdf)
[FigS6C_other-solid-tumors-barplot.pdf](https://github.com/user-attachments/files/25134287/FigS6C_other-solid-tumors-barplot.pdf)

Old version: 
https://github.com/AlexsLemonade/scpca-paper-figures/blob/main/figures/compiled_figures/pngs/figure_s6.png


sjspielman (commented):
Yay more cells, and yay no new colors!

This all looks good, but you'll need to fix the sarcoma legend text - HPSC's are cut off on the right.

allyhawkins:
> This all looks good, but you'll need to fix the sarcoma legend text - HPSC's are cut off on the right.

Good catch! I fixed this in https://github.com/AlexsLemonade/scpca-paper-figures/pull/400/commits/a2f4d29f7d4ce6b990b67e9540c128787120cccf

sjspielman (approved):
👍

PR #402, opened 2026-02-06
New Panels for S4

### LINKED ISSUE #386: Regenerate panels for S4
We should regenerate the panels for S4 to incorporate new consensus cell types.

allyhawkins:
Closes #386 

I updated the panels for S4 by downloading the updated files on staging and then re-running the script. There was a tie in one of the top consensus T cell types with both having n=93 cells. I updated the code to just take the first cell type in the case of the tie, which I think is fine to keep the number of panels even. There were some very minor cell type number changes, but otherwise the same.


sjspielman (approved):
> There was a tie in one of the top consensus T cell types with both having n=93 cells. I updated the code to just take the first cell type in the case of the tie, which I think is fine to keep the number of panels even. 

This approach seems fine to me, and (unless you disagree?) I don't think we need to mention handling ties in the legend since it doesn't really change the outcome of the figure, and code is all public!

allyhawkins:
> > There was a tie in one of the top consensus T cell types with both having n=93 cells. I updated the code to just take the first cell type in the case of the tie, which I think is fine to keep the number of panels even.
> 
> This approach seems fine to me, and (unless you disagree?) I don't think we need to mention handling ties in the legend since it doesn't really change the outcome of the figure, and code is all public!

Yes definitely agree with this!

PR #408, opened 2026-02-06
Update pseudobulk analysis and results

### LINKED ISSUE #404: Update ORA code for pseudobulk analysis
Broken out of https://github.com/AlexsLemonade/scpca-paper-figures/issues/393

This issue tracks updating the pseudobulk analysis to use panglao gene sets directly for the ORA. Changes will include:
- deprecating (but not removing; I will just add comments it's no longer used) the script that previously created the gene sets
- updating the ORA notebook to use the Panglao gene sets
- associated updates to the bash run script 

I plan to include the regenerated notebook files in this PR, as well.

sjspielman:
Closes #404 
This PR updates the pseudobulk analysis. There's a lot of file changes, but it's not as much as it looks! Here's a rundown:

## Code changes (most of review)

- I added comments to deprecate the previous gene set creation script, and updated the scripts/readme as well to this effect
- I updated the notebook that runs ORA to read in the Panglao gene set files directly, via a new parameter passing in this file
- I updated the `run-prediction.sh` script accommodate these changes
  - While I was here, I also simplified it - we had been looping over different expression thresholds (used for determining which genes to include in models), but only the value `0` is used in the paper. So, rather than looping, I made this an argument that defaults to what we actually use in the paper
  - I also added a step to run the exploratory notebook that plots the data (and happily I didn't have to change any text in it!), doesn't hurt!

## Notebook changes

The following notebooks got re-rendered:

- the model notebook at the 0 expression threshold specifically. See it here: [build-assess-models_threshold-0.nb.html](https://github.com/user-attachments/files/25139354/build-assess-models_threshold-0.nb.html)
- the ORA notebooks, 1 per project. See one of them here if you'd like: [perform-ora_SCPCP000001.nb.html](https://github.com/user-attachments/files/25139239/perform-ora_SCPCP000001.nb.html)
- the exploratory view-ora-results notebook. See it here: [view-ora-results.nb.html](https://github.com/user-attachments/files/25139070/view-ora-results.nb.html)

## Other file changes

- I removed the previous geneset files. I updated data/readme here too.
- The result TSVs with ORAs themselves got updated


allyhawkins (approved):
LGTM

PR #409, opened 2026-02-09
Recompile figure 5

### LINKED ISSUE #390: Recompile Figure 5
We will need to recompile Figure 5 with the updated panels

sjspielman:
Closes #390 
In addition to the compilation, I removed an old pdf of panel 5B - this isn't used since it proved too big. We only use png now, see code -
https://github.com/AlexsLemonade/scpca-paper-figures/blob/da4dadc63ce45a2e300525c7e66f747f0085c75c/scripts/Fig5A-B_nb-umap-panels.R#L45-L47


allyhawkins (approved):
LGTM

PR #410, opened 2026-02-09
Update pseudobulk analysis panels

### LINKED ISSUE #405: Regenerate pseudobulk figures
Broken out of https://github.com/AlexsLemonade/scpca-paper-figures/issues/393

Following #404, we will need to regenerate figure panels for Figures 6 and S7.

sjspielman:
Closes #405 
Stacked on #408 

This PR regenerates the pseudobulk figures. There's a couple changes here:

- Since the cell type names are now way longer, the layout we had before was a little tricky. So, I switched it to a vertical layout with horizontal barplots which seems to be the move for these cell type names! I updated various bits of plotting code to make this happen including resizing.
- I ended up removing the `geom_text()` because it was making the plot way too busy, now that we have more cell types coming back significant from the ORA. All the odds ratios are in this repo as TSVs, so I think this is fine, but if we are concerned about removing that information we could always add a Table S3. Don't think it's needed but let me know if you disagree.
- Finally, we had been tracking the number of gene sets tested for each project, which is no longer necessary. I went ahead and removed that aspect of the code (and therefore from the numbers tsv itself). Note that I've updated the manuscript to the panglao number of cell types here: https://github.com/AlexsLemonade/ScPCA-manuscript/pull/261/changes/c3e8ae35e451e11e698feb172be46c823442ff68

If we're on board with these figure changes, the next step of compilation will just be adding `A.` and `B.` and inserting just a a bit of spacing between A/B so it's not so tight.


manuscript-numbers/bulk-analysis-counts.tsv
```diff
@@ -1,7 +1,7 @@
-scpca_project_id	n_samples_total	n_samples_low_quality	n_samples_used	n_ora_genesets
-SCPCP000001	18	1	17	36
-SCPCP000002	20	2	18	36
-SCPCP000006	39	4	35	31
-SCPCP000009	4	0	4	36
-SCPCP000017	24	1	23	37
-total	105	8	97	NA
+scpca_project_id	n_samples_total	n_samples_low_quality	n_samples_used
```
allyhawkins: I think we still want to have the number of genesets used? It's still dependent on the panglao reference used since the number of cell types is different for each reference.

scripts/Fig6-FigS7_bulk-analysis.R
```diff
@@ -4,7 +4,15 @@ renv::load()
...
 # load any libraries 
 library(ggplot2)
-theme_set(theme_bw())
+theme_set(
+  theme_bw() + 
+    theme(      
+      legend.text = element_text(size = 8), 
+      legend.title = element_text(size = 9), 
+      axis.text = element_text(size = 9),
+      legend.position = "bottom"
```
allyhawkins: Something is weird with both legends where the text is not centered with the legend bar. Can we fix that here or will we need to do that in illustrator?

scripts/Fig6-FigS7_bulk-analysis.R
```diff
@@ -87,27 +85,20 @@ plot_scatterplot <- function(df) {
...
   ggplot(df) +
     aes(
-      x = tidytext::reorder_within(geneset_cell_type, odds_ratio, within = project_facet),
-      y = odds_ratio, 
+      x = odds_ratio, 
+      y = tidytext::reorder_within(geneset_cell_type, odds_ratio, within = project_facet),
       fill = -log10(p_adj_bh)
     ) +
     geom_col() +
-    geom_text(
```
allyhawkins, line 95: How busy does this make it? I think if we can show the odds ratios we should...

allyhawkins (commented):
I like this new layout, but I had a few comments about things we might want to keep that you removed. I think we do want to report the number of genesets, since that's dependent on the reference used. I also think it's worth showing the odds ratios if we can.

scripts/Fig6-FigS7_bulk-analysis.R
sjspielman: Let me (quickly) see if I can fix this - if not, Illustrator!

manuscript-numbers/bulk-analysis-counts.tsv
sjspielman: I can update to record those numbers instead, code changes incoming.

scripts/Fig6-FigS7_bulk-analysis.R
sjspielman: Screenshot 2026-02-09 at 12 19 30 PM

here's a screenshot with ORs

scripts/Fig6-FigS7_bulk-analysis.R
sjspielman: I moved the legend title to the top of the legends, looks ok to me now. What do you think? I can also center the legend title (it's left-justified right now) if you think that will look better?
Screenshot 2026-02-09 at 12 39 53 PM

allyhawkins: I like the legend title placement here!

scripts/Fig6-FigS7_bulk-analysis.R
allyhawkins: I think I would keep the ratios here. It's busy yes, but still legible.

sjspielman:
Odds ratios restored, with a few small code changes to change sizing between main and SI figures to fit.

allyhawkins (approved):
LGTM

PR #414, opened 2026-05-08
added pdf, png, and svg of graphical abstract

dvenprasad:
Added the graphical abstract files to relevant directories


jaclyn-taroni (approved):
🚀

PR #416, opened 2026-05-10
Rework HGG/LGG classification

jaclyn-taroni:
On Friday, the pLGG submitters mentioned that they did not include the ependymoma (EPN) sample in their pLGG paper because it was an ependymoma sample. That made me wonder if we included it in the pLGG grouping for Figure 4B and 4C simply because it was in that project; it looks like we did. Here, I'm removing the EPN sample from the pLGG panels and moving the PXA sample to pLGG, as those are typically considered low-grade gliomas (and are typically grade 2). (I expect that particular PXA might be in the pHGG project because it is secondary to a GBM.) There is no change to interpretation, in my opinion, as this should only impact 2 samples total. (This is quite a descriptive figure in the first place.)

To recreate this figure, I followed all the steps to download data here: https://github.com/AlexsLemonade/scpca-paper-figures#instructions-to-prepare-data

Then, I updated `scripts/figure_setup/assign-brain-classifications.R` and regenerated the bar plots by running `Fig4B-C_brain-barplots.R`. I recompiled the PDF of the figure in Illustrator and then exported to PNG.

Questions:

- Did I miss any steps here? That includes regenerating other figures that depend on this file, but I didn't find any.
- The PNG I exported is much larger than the last one. Is there a particular process we follow for exporting?


allyhawkins (approved):
This all looks good to me, and I'm on board with moving the PXA sample too. As for exporting the png, we don't have a particular process other than just using the export option in illustrator. There might be some underlying setting that's different in your illustrator (or one of ours), but I think if the content is okay then it should be good to keep as is.

